PR #1, opened 2018-01-10
[WIP] Steps for adding the aerogearcatalog apbs to an openshift cluster

david-martin:
I already had the aerogearcatalog org in my config (as I'm using the mobiel-core cluster).
So for verifying, I modified it to point to my own dockerhub org i.e. davidmartin.
After doing the relist, I could only see apbs in my org.

### DIFF
--- a/adding-services-to-ansible-broker.adoc
+++ b/adding-services-to-ansible-broker.adoc
@@ -20,13 +20,15 @@ Add the following to the `registry` array:
 ----
 registry:
   - type: "dockerhub"
-    name: "dh"
+    name: "aerogearcatalog"
     url: "https://registry.hub.docker.com"
     org: "aerogearcatalog"
     tag: "latest"
     white_list:
       - ".*-apb$"
 ----
+
+NOTE: The `name` must be different for each registry entry.
 
 == Redeploy the Broker
 
--- a/adding-services-to-ansible-broker.adoc
+++ b/adding-services-to-ansible-broker.adoc
@@ -1,0 +1,71 @@
+= Adding Mobile Services to the OpenShift Ansible Broker
+
+== Prerequesites
+
+* OpenShift Origin >= 3.7
+** https://docs.openshift.org/latest/install_config/index.html
+* OpenShift Ansible Broker
+** https://docs.openshift.org/latest/install_config/install/advanced_install.html#configuring-openshift-ansible-broker
+
+== Reconfigure the Broker
+
+[source,bash]
+----
+oc edit configmap broker-config -n ansible-service-broker
+----
+
+Add the following to the `registry` array:
+
+[source,yaml]
+----
+registry:
+  - type: "dockerhub"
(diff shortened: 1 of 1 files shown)


adding-services-to-ansible-broker.adoc
```diff
@@ -0,0 +1,71 @@
...
+oc edit configmap broker-config -n ansible-service-broker
+----
+
+Add the following to the `registry` array:
+
+[source,yaml]
+----
+registry:
+  - type: "dockerhub"
+    name: "dh"
```
pb82: The name must be unique, maybe it's worth pointing that out (in case users already have other registries installed)

adding-services-to-ansible-broker.adoc
```diff
@@ -0,0 +1,71 @@
...
+[source,bash]
+----
+docker run --rm --privileged -v $HOME/.kube:/.kube -v /var/run/docker.sock:/var/run/docker.sock -u 1000 docker.io/ansibleplaybookbundle/apb-tools:latest relist
+----
+
+NOTE: You will need to be logged into your OpenShift cluster with a user who has the necessary permissions to relist the catalog. The `cluster-admin` role should have all necessary permissions.
+
+== Verify Mobile Services are shown in the Catalog
+
+In the OpenShift Catalog, additional Services should now be shown. These will appear either in the 'Other' category, or in a 'Mobile' category, if that is available in your OpenShift cluster.
```
pb82: As far as i know the `Mobile` category will be shown automatically because the APBs in aerogearcatalog are tagged with `Mobile`. Or is any other config required to make that category visible?

david-martin: See https://github.com/aerogear/mobile-core/blob/master/ui/public/scripts/app.js#L13-L30 in our UI extension where we add the Mobile category, and Apps & Services sub categories.
By default, the Catalog doesn't have these categories.

So those categories only appear if:
* our UI extension is enabled in the OpenShift cluster (that bit of js is loaded in the browser)
* at least 1 catalog item that fits into the category is available

adding-services-to-ansible-broker.adoc
david-martin: Good point, will add some text

david-martin: See https://github.com/aerogear/service-docs/pull/1/commits/089116753f7794d68dd78d22bdfcb31406a08520

pb82: 👍

adding-services-to-ansible-broker.adoc
pb82: Good to know, thanks!

david-martin:
@philbrookes @pb82 Are you happy to merge?

pb82:
@david-martin yes, looks good

philbrookes:
@david-martin yes, looks good to me.

PR #2, opened 2018-01-11
Docs for making mobile services discovered by metrics

aliok:
Just added the docs. We can think about the structure later when we have more docs.
We should discuss with the docs team at that point as they have some awesome ideas.

cc @wtrocki @darahayes @david-martin (random people from the team): appreciate the early feedback
Can't add "reviewers" on Github.

### DIFF
--- a/making-mobile-services-discoverable-by-metrics.adoc
+++ b/making-mobile-services-discoverable-by-metrics.adoc
@@ -70,7 +70,7 @@ oc annotate service my-awesome-service "org.aerogear.metrics/plain_endpoint"="/m
 oc annotate service my-awesome-service "org.aerogear.metrics/ssl_endpoint"="/metrics"
 ----
 
-Prometheus constantly monitors Kubernetes and will be notified about this change. It will start getting the metrics
+Prometheus constantly monitors Kubernetes and will be notified about this change. It will start collecting the metrics
 immediately.
 
 == How auto-discovery works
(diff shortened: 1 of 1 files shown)


aliok:
UPDATE: just saw the comments in https://gist.github.com/aliok/956d814756bd53d09cf704790f9710e2
I will update the PR

aliok:
Pushed some changes. Changed the PR to WIP as I have some stuff to verify first

making-mobile-services-discoverable-by-metrics.adoc
```diff
@@ -0,0 +1,58 @@
...
+    name: fh-sync-server
+    annotations:
+      org.aerogear.metrics/plain_endpoint: "/metrics"
+    labels:
+      ...
+  spec:
+    ...
+----
+
+TODO: will document `oc patch`
```
aliok: I will write this soon

making-mobile-services-discoverable-by-metrics.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+[source,yaml]
+----
+org.aerogear.metrics/ssl_endpoint: "/rest/metrics"
+----
+
+This endpoint mustn't have any authentication currently. This is subject to change in the future.
+
+Also, the endpoint should be accessible by the Prometheus instance. There should be a route for the endpoint
+that Prometheus can access. This means, service should be in the same project as Prometheus.
```
aliok: @david-martin I wrote this sentence according to your comment on the Gist https://gist.github.com/aliok/956d814756bd53d09cf704790f9710e2, but I seconded that the service must be in the same service as Prometheus instance. Is this because of Prometheus' Kubernetes SD limitation?
I am asking because the route doesn't matter here: route could be made accessible by any service on any project.

darahayes:
@aliok This is great. I have one question. Would it be possible to get a short explanation of how the service discovery mechanism works in Prometheus? Glancing at the Prometheus docs, there are many types of service discovery. It would be nice if we made it obvious in this doc which mechanism we are using and perhaps a link to it in the prometheus docs.

making-mobile-services-discoverable-by-metrics.adoc
david-martin: @aliok The endpoint annotation will determine just the endpoint e.g. `/metrics`
The host will be set to the Kbuernetes DNS name of the service, which will only be accessible within the same project as Prometheus, unless project networks are linked.
More info in the docs https://prometheus.io/docs/prometheus/latest/configuration/configuration/#service

Might be good to link to that docs page to say thats what's being used in the prometheus config to discover services (as per @darahayes suggestion)

aliok:
@darahayes makes sense. will do :)

making-mobile-services-discoverable-by-metrics.adoc
aliok: > Kbuernetes DNS name of the service
this is the key: not the public route; K8s DNS name. 
ok, got it now. thanks @david-martin

aliok:
@darahayes as we talked today during the daily standup, we won't add detailed information about how this works here. Target of this documentation is basically our community and possibly customers, not RHMAP team.
I added some docs and comments on metrics-apb directly: https://github.com/aerogearcatalog/metrics-apb/pull/13

aliok:
This PR is no more a "[WIP]".
cc @david-martin @darahayes @wtrocki

PS: @david-martin still can't add reviewers over Github UI :(

darahayes:
Sounds good @aliok 👍

david-martin:
@aliok I changed the perm on this repo to 'Write' for Pushers. That might help?

PR #4, opened 2018-01-11
[AEROGEAR-1802] Add manual service addition to metrics service guide

paolobueno:
- [X] Add web-based instructions
- [X] Add cli-based instructions
- [x] Add prometheus screenshots

NOTE: depends on https://github.com/aerogearcatalog/metrics-apb/pull/14


manually-adding-services-to-metrics.adoc
```diff
@@ -0,0 +1,120 @@
...
+    scrape_configs:
+      - job_name: prometheus
+        static_configs:
+          - targets:
+            - "localhost:9090"
+      - job_name: grafana
+        scheme: https
+        static_configs:
+          - targets:
+            - "grafana.myproject.svc:4181"
```
paolobueno: @david-martin I'm having trouble make the grafana endpoint actually work because it's behind the openshift auth proxy.

I was wondering if it would be relevant to actually use @aliok 's [sample node app](https://github.com/aliok/node-app-with-prometheus-metrics) instead and if we should be looking into exposing the grafana metrics endpoint without auth.

IIUC we created the prometheus-internal service to enable grafana to reach it without going through the proxy, but is there a way to configure the proxy to let requests to those specific endpoints (/metrics and whatever grafana has to query prometheus for datases) through?

cc @darahayes

darahayes: @paolobueno If my understanding is correct this document is showing an example where prometheus is scraping metrics out of the existing Grafana service is that correct?

Perhaps there is a way to allow specific requests through the proxy without auth. Whether it's based on endpoint or origin or something else. However I don't see this as the best approach. In this scenario Prometheus would be talking to Grafana through its route as defined in OpenShift.

It's my understanding that OpenShift routes are used to publicly expose services on the internet in a production scenario. I may be wrong but I think the convention in Kubernetes/Openshift is that applications in the same namespace should be talking to each other using a Service. Similar to the prometheus-internal service.

paolobueno: @darahayes Yes, I've only picked grafana as the example target in the context of the guide because 1. it's not currently labelled for discovery by prometheus, so manual addition would be required and 2. it's already provisioned alongside it, so there wouldn't be a prereq for another apb.

Ah yes, the 4181 port there is wrong, forgot to update. The grafana service exposes the default https 443 instead, while 4181 is indeed internal to the pod.

I won't update it right now not to flag this thread as out-of-date, but my take is that this should be:

```yaml
- job_name: grafana
  scheme: https
  tls_config:
    insecure_skip_verify: true
  static_configs:
    - targets:
      - "grafana:443"
```

Acessing grafana either through the service binding or even from the public route (https://grafana-myproject.192.168.37.1.nip.io/metrics) still go through the proxy and the response is 403 since the prometheus client isn't authenticated, so that's where I'm stuck.

david-martin: @paolobueno You can use thie `-skip-auth-regex` flag in the oauth proxy
```
value: bypass authentication for requests path's that match (may be given multiple times). Cannot be set with -bypass-auth-except-for
```
https://github.com/openshift/oauth-proxy#command-line-options

darahayes: @paolobueno yes the two options are:

1. Prometheus talks to Grafana through the oauth proxy. This means figuring out how to allow the requests through the proxy. It looks like @david-martin's suggestion might help with that.

2. Create a new service that directly exposes Grafana within the namespace.

There are benefits and drawbacks to each approach I guess.

paolobueno: Thanks a lot for the help. I guess creating an internal service for grafana too might be the best way to go, since exposing the metrics endpoints to the outside unconditionally might have security implications... going to move forward on this for now.

paolobueno:
@aliok thank you! Yes, this should now be ready for verification, it references a new `grafana-internal` service definition available [here](https://github.com/aerogearcatalog/metrics-apb/pull/14), but the goal is to be generic enough to apply to other services, of course.

manually-adding-services-to-metrics.adoc
pb82: +1 for providinig an example using an internal service to access to grafana metrics endpoint

manually-adding-services-to-metrics.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+Run the following commands:
+
+[source,bash]
+----
+oc login # login to the OpenShift Cluster
+oc get projects # list available Projects
+oc use myproject # switch to the project where the prometheus and the custom service are located
+oc get configmaps # list available Config Maps
+oc edit configmap prometheus-configmap # open the prometheus configmap in $OC_EDITOR or $EDITOR
```
paolobueno: As a side note I've had some troubles with `oc edit`-ing the prometheus configmap due to linebreaks coming up as `\n`

aliok:
@paolobueno merging this to start working on https://issues.jboss.org/browse/AEROGEAR-1797

PR #8, opened 2018-02-02
Document topic on 'working with grafana dashboards'

darahayes:
Original Ticket: https://issues.jboss.org/browse/AEROGEAR-1923

Please note that although the original ticket mentions 'Documenting how to add dashboards via a config map' This is actually something that an end user should not be doing. As such it makes no sense to document it. Instead this document gives the end-user directions on how dashboards should be added, modified, imported exported. It also instructs the end-user what they should do if they want to modify the default dashboards provided with each AeroGear service.

ping @finp


finp:
@darahayes  I've made some edits, please push back if I've gone in the wrong direction

PR #12, opened 2018-02-08
Mobile services dashboard documentation

psturc:
https://issues.jboss.org/browse/AEROGEAR-1961

This doc will probably need better and more detailed explanation of **Active Requests** and **Lag of Event Loop** values. Any suggestions?


services/mobile-services-dashboard.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+If you want the Keycloak service's metrics data to be available for Graphana, you first need to deploy it
+from the Service Catalog: https://github.com/aerogearcatalog/keycloak-apb
+
+*Note:* Keycloak service must be provisioned within the same namespace,
+otherwise the Keycloak's metrics endpoint won't be reachable for Prometheus and no data will be displayed.
+
+=== Version
+
+Shows the information about the version of the Node.js
```
josemigallas: Ideally this would be the service's version or if this is not possible the **java** version.

services/mobile-services-dashboard.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+If you want the Sync service's metrics data to be available for Graphana, you first need to deploy it
+from the Service Catalog: https://github.com/aerogearcatalog/fh-sync-server-apb
+
+*Note:* Sync service must be provisioned within the same namespace,
+otherwise the Sync metrics endpoint won't be reachable for Prometheus and no data will be displayed.
+
+=== Version
+
+Shows the version number of the Keycloak service.
```
josemigallas: Keycloak -> Sync :)

services/mobile-services-dashboard.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+=== Keycloak service description
+
+This panel just gives you basic information about the Keycloak service
+
+=== Provisioned
+
+This panel informs you about the status of connection with the Keycloak service.
+
+If you want the Keycloak service's metrics data to be available for Graphana, you first need to deploy it
```
StevenTobin: Graphana -> Grafana

services/mobile-services-dashboard.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+=== Sync service description
+
+This panel just gives you basic information about the Sync service
+
+=== Provisioned
+
+This panel informs you about the status of connection with the Sync service.
+
+If you want the Sync service's metrics data to be available for Graphana, you first need to deploy it
```
StevenTobin: Graphana -> Grafana

services/mobile-services-dashboard.adoc
pb82: @psturc Node.js -> Java @josemigallas how do you get the KeyCloak version?

services/mobile-services-dashboard.adoc
pb82: Copy Pasta 🍝

services/mobile-services-dashboard.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+=== Version
+
+Shows the version number of the Keycloak service.
+
+=== Active Requests
+
+Number of active requests.
+
+=== Lag of Event Loop
```
pb82: Event Loop -> Sync Loop

services/mobile-services-dashboard.adoc
josemigallas: @pb82 That's to be figured out yet

pb82: @josemigallas one option would be to reda the `jave.version` system property in the SPI: https://stackoverflow.com/questions/2591083/getting-java-version-at-runtime

josemigallas: Well sure, the Java version is what's currently being displayed. There is actually a metric for that already provided. What I mean is that it would be more useful to now what version of the service itself (APB I guess) is deployed, wouldn't it?

david-martin:
cc @finp.
This seems like a 'reference' doc currently.
I'm wondering if its possible to change this around to be more 'action' based i.e. what's the objective for someone viewing this doc?

@StevenTobin Do you think this is adding enough value besides what help/text is available in the dashboard itself?

StevenTobin:
@david-martin this document could be made more action oriented by changing it to describe the use of the overview dashboard. That the overview will not be populated at first but will as services are provisioned, this could logically include an overview of the panels describing what to expect when the service is provisioned.

We don't call out specifically in the dashboard what each panel metric describes, although they are fairly self-explanatory, I don't know what we should be assuming the user knows and needs to be told.

finp:
I'm thinking that this could be improved as follows:
* split in two, generic and specific
** "Monitoring mobile services using the Grafana Mobile Service Dashboard"
** make second part specific, 'Monitoring the Keycloak mobile service'
* we can add other services in time
* The goal here is for users to be able to monitor and react, what if no. of log in errors is large, is there something you should check?
* +1 to showing service version - this would really help users navigate the docs (I'm guessing that after installing,  users won't know what version they're running, and this would provide that info)
@psturc  want me to edit this?

psturc:
@finp yes, I think it would be better if you could directly edit this. Thanks.

Also we have some stuff for Keycloak already done https://github.com/aerogear/mobile-docs/blob/master/services/keycloak-dashboard.adoc

PR #18, opened 2018-04-12
docs: guides for integration between android app and custom service

darahayes:
Associated JIRA Ticket: https://issues.jboss.org/browse/AEROGEAR-2435

Pinging @finp because 

1. I have no idea where the appropriate place for these docs are
2. I'm not sure about the structure/formatting


services/integrate-android-app-with-custom-service.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+== Integrate your Android Application with the Custom Service
+
+Take a copy of the JSON configuration from the mobile client and save it in `app/src/main/assets/mobile-services.json` in your Android application's source code.
+
+Now the Aerogear SDK can be used to get the configuration for your custom service and it is possible to make requests. Please look at the example code below:
+
+[source,java]
+---
+// code example goes here once I can collaborate with Ali
```
darahayes: @aliok Perhaps you could help me put a code example together for getting the custom service config and then making a HTTP request?

aliok: @darahayes 

This is subject to change because of the changes in the core side. But, here's the current procedure:

```
// get the service URL
final ServiceConfiguration serviceConfiguration = mobileCore.getServiceConfiguration("custom-runtime-connector");
final String serviceUrl = serviceConfiguration.getUrl();

// make the request
HttpRequest httpRequest = mobileCore.getHttpLayer().newRequest();
httpRequest.post(serviceUrl, "THE PAYLOAD".getBytes("UTF-8"));
HttpResponse httpResponse = httpRequest.execute();

httpResponse.onError(() -> {
    // handle error
});
httpResponse.onSuccess(() -> {
    // handle success
});
httpResponse.onComplete(() -> {
    // called regardless success or error
});
```


How we are getting the service config is going to change a bit.

aliok:
@darahayes docs look good to me. read it through. great work.

darahayes:
@aliok thanks a million for the code example. I've updated the PR with it!

services/integrate-android-app-with-custom-service.adoc
```diff
@@ -0,0 +1,70 @@
+= Integrate an Android Application with Custom Services Running in OpenShift
```
finp: Creating an Android App that calls an AeroGear Mobile Custom Service?
@darahayes  trying to frame this in a context that a user would want to achieve

darahayes: @finp I'm not a fan of the title myself but it wouldn't be right to say 'AeroGear Mobile Custom Service.' 

In laymans terms this document shows 'How to connect my app to my backend service using the Aerogear SDK'

We're talking about how we can integrate an Android App to a backend service written, maintained and owned by the end user such that they can access this service seamlessly using the Aerogear Android SDK.

I think 'AeroGear Mobile Custom Service.' is not the right terminology because the service has nothing to do with Aerogear and the service isn't necessarily mobile focused.

finp: k, let me digest that, thx ;)

services/integrate-android-app-with-custom-service.adoc
```diff
@@ -48,6 +48,23 @@ Now the Aerogear SDK can be used to get the configuration for your custom servic
 
 [source,java]
 ---
-// code example goes here once I can collaborate with Ali
+// get the service URL
```
finp: @aliok  Did this example come from an existing code? or is it a standalone example for this doc? 
reason I ask is that we could remote include tagged code if it makes sense in this case (and never have to worry about it becoming stale)

darahayes: @finp From what I can tell, it's just a generic example.

services/integrate-android-app-with-custom-service.adoc
```diff
@@ -0,0 +1,70 @@
...
+
+== Prerequisites
+
+* link:./create-android-client-on-openshift.adoc[Create an Android Mobile Client in OpenShift]
+
+== Procedure
+
+For demonstration, this guide uses the example Python application which can be found in the OpenShift Service Catalog. The same steps can be followed any HTTP web service running in OpenShift that has a publicly accessible route.
+
+image:./images/custom-service-py.png[Custom Service and Android Client in OpenShift]
```
david-martin: This image looks out of place here. Should it be at the end of the next section?

services/integrate-android-app-with-custom-service.adoc
```diff
@@ -0,0 +1,70 @@
...
+image:./images/mobile-client-custom-runtime.png[Mobile Client Config with Custom Service]
+
+== Integrate your Android Application with the Custom Service
+
+Take a copy of the JSON configuration from the mobile client and save it in `app/src/main/assets/mobile-services.json` in your Android application's source code.
+
+Now the Aerogear SDK can be used to get the configuration for your custom service and it is possible to make requests. Please look at the example code below:
+
+[source,java]
+---
```
david-martin: The source formatting looks off here in the GH preview

services/integrate-android-app-with-custom-service.adoc
darahayes: I guess the idea with that image was to show that we've already got a mobile client and a web service of some sort. Perhaps it's not the best though

services/integrate-android-app-with-custom-service.adoc
```diff
@@ -0,0 +1,72 @@
...
+* In the Service Config field, place any additional configuration that might be needed by the mobile app in order to communicate with the backend service. **This must be a valid JSON format**
+* Click 'Create'.
+
+image:./images/create-crc.png[Provisioning a Custom Runtime Connector]
+
+Note that provisioning the Custom Runtime Connector simply creates a Config Map that is used to populate the `mobile-services.json` file which is consumed by your mobile application. It does not deploy any new services.
+
+=== Add the Custom Runtime Connector Service to the Mobile Client
+
+* Click the Mobile Client to reveal the JSON configuration. The 'services' list is empty.
```
david-martin: The services list was *not* empty for me. The custom runtime connector config was already there

darahayes: This is what the expected default behaviour will be once the changes to the CRC mentioned by @philbrookes are merged.

darahayes: It's not currently the case but it will be in the next week or so right?

services/integrate-android-app-with-custom-service.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+[source,java]
+--
+// get the service URL
+final ServiceConfiguration serviceConfiguration = mobileCore.getServiceConfiguration("custom-runtime-connector");
+final String serviceUrl = serviceConfiguration.getUrl();
+
+// make the request
+HttpRequest httpRequest = mobileCore.getHttpLayer().newRequest();
+httpRequest.post(serviceUrl, "THE PAYLOAD".getBytes("UTF-8"));
```
david-martin: This code worked :+1: 
2 minor things, probably no action needed:

* I did have to wrap this line in a try/catch.
* This http POST call gave a 403 when I used the Python app (as recommended in the guide). I don't think there's an API server with sample endpoints available, but maybe we could change this example to a GET and hit the 'health' endpoint i.e. `http://pyapp-myproject.192.168.99.100.nip.io/health` ?

darahayes: @david-martin trying to progress this and close it out. I'm not actually familiar with using the SDK. Which particular bit did you need to wrap with a try/catch? The `http.post` bit? 

Why would a `http.post` require a try/catch but a `http.get` wouldn't? It might be misguided but to me, the pattern of using the onSuccess, onError and onComplete handlers would suggest that any exceptions would be caught internally and surfaced in the error handler, regardless of the HTTP method.

Or is it a case that the onError handler only takes care of application layer errors e.g. HTTP 4xx-5xx responses and we must wrap the whole thing in try/catch?

@danielpassos would you be know what's going on here?

darahayes:
I've incorporated @david-martin's feedback around image placement and the code example. I've also fixed up some punctuation and headers.

@finp From your point of view is there anything else that needs to be done to progress this? For example, are the docs actually in the right location?

services/integrate-android-app-with-custom-service.adoc
david-martin: The call to `.getBytes("UTF-8")` throws an exception.

danielpassos: Hey @darahayes Sorry, I think I'm late, @david-martin have already answered that. The problem is [String.html#getBytes](https://developer.android.com/reference/java/lang/String.html#getBytes(java.lang.String))

finp:
merging, but this doc doesn't reflect the current ui, and won't be presented to website until the ui code changes are merged

PR #29, opened 2018-06-08
Minor cleanup of Readme and scripts

josemigallas:
Motivation: https://issues.jboss.org/browse/AEROGEAR-2990

As result of that JIRA, I found a couple of minor details to be fixed. They are described as commit messages. No errors or wrong information whatsoever.


README.adoc
```diff
@@ -104,11 +104,7 @@ https://docs.antora.org/antora/1.0/install/troubleshoot-nodegit/
 ./bin/build.sh local-site.yml
 ----
 + 
-After building for the first time, you can use the following command for subsequent builds:
-+
-----
-./bin/build.sh local-site.yml
```
josemigallas: @finp I removed this part because the command line is the same. Maybe it was only a typo and it was intended to be `quick-build.sh` instead of `build.sh`?

finp:
@josemigallas  that's it, good spot, can you change it to quick-build.sh, and then we're good to merge

PR #33, opened 2018-06-08
Small changes to "Sending Push Notifications"

josemigallas:
Motivation: https://issues.jboss.org/browse/AEROGEAR-3004

This document seems to be very concise and straightforward and I don't think any screenshot is needed.

On the other hand, I wonder if more info about the filtering should be added _into another page_.


pb82:
@josemigallas @jstaffor On the top of this file there is an error: `Unresolved directive in <stdin> - include:{partialsdir}/attributes.adoc[]`

PR #43, opened 2018-06-12
Handling a Message, with code examples from showcase apps

josemigallas:
Motivation: https://issues.jboss.org/browse/AEROGEAR-3007

There are many problems with tagging the code in showcase examples.

- In Android, the code is in a XML file so the comments have a different format than `//`.
- In Cordova, the implementation is very attached to the framework and architecture of the showcase app (ionic) and it's not possible to keep the snippet brief and clear (see [code](https://github.com/aerogear/cordova-showcase-template/blob/master/src/services/push.service.ts)).

### DIFF
--- a/modules/ROOT/pages/push/handling-push-notifications.adoc
+++ b/modules/ROOT/pages/push/handling-push-notifications.adoc
@@ -50,6 +50,5 @@ The {SDK} for iOS does not provide a way to handle incoming messages. Instead, f
 ****
 
 The {SDK} for JavaScript does not provide a way to handle incoming messages. We recommend using the link:https://ionicframework.com/docs/native/push[Ionic Push plugin].
-[source,javascript]
 
 ****
--- a/modules/ROOT/pages/push/handling-push-notifications.adoc
+++ b/modules/ROOT/pages/push/handling-push-notifications.adoc
@@ -1,3 +1,54 @@
 include::{partialsdir}/attributes.adoc[]
 
 = Handling Push Notifications
+
+Follow the next steps to handle incoming {push-notification}s in your foregrounded application.
+
+NOTE: {push-notification}s that arrive when the application is in the background will always be handled by the OS.
+
+[role="primary"]
+.Android
+****
+Add your implementation of *MessageHandler* to your application's manifest.
+
+[source,xml]
+----
+<meta-data
+    android:name="DEFAULT_MESSAGE_HANDLER_KEY"
+    android:value=".MyMessageHandler" />
+----
+****
+
+
+[role="secondary"]
+.iOS
+****
+[source,swift]
+----
+TODO
+----
+****
+
+
+[role="secondary"]
+.Cordova
+****
+[source,javascript]
+----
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/push/handling-push-notifications.adoc
```diff
@@ -1,3 +1,62 @@
...
+----
+<meta-data
+    android:name="DEFAULT_MESSAGE_HANDLER_KEY"
+    android:value=".MyMessageHandler" />
+----
+****
+
+
+[role="secondary"]
+.iOS
```
josemigallas: @jhellar could you evaluate if this is correct? I could use some help here.

jhellar: I think that you first need to handle remote notification with https://github.com/aerogear/ios-showcase-template/blob/master/ios-showcase-template/AppDelegate.swift#L116. NotificationCenter is not part of AGSPush. It can be used to post the notification (https://github.com/aerogear/ios-showcase-template/blob/master/ios-showcase-template/push/PushHelper.swift#L18-L20) to a View so that it can be rendered. @wtrocki can you please confirm?

finp: I agree with @wtrocki  that this is too big to take on, and we should skip it.. 
However, I disagree with the general point. It's ok to document any steps that user needs to perform in order to be successful with mobile services regardless of software origin...  provided that documentation task is reasonable (and in this case, I don't think it is)

wtrocki: @finp problem here is that this code will outdate faster than yoghurt in the fridge and people will actually get better experience if we point them to: 
https://developer.apple.com/notifications/

modules/ROOT/pages/push/handling-push-notifications.adoc
```diff
@@ -1,3 +1,62 @@
...
+const pushObject = new Push().init({
+    android: {},
+    ios: {
+      alert: true,
+      badge: true,
+      sound: true
+    }
+});
+
+pushObject.on("notification")
```
wtrocki: This is specific to ionic extension. Worth to mention that.

josemigallas:
OK I updated it. I am in favor of not describing parts that are out of our SDK, it's too volatile and actually up to each particular user.

modules/ROOT/pages/push/handling-push-notifications.adoc
```diff
@@ -1,3 +1,55 @@
...
+
+****
+
+
+[role="secondary"]
+.Cordova
+****
+
+The {SDK} for JavaScript does not provide a way to handle incoming messages. We recommend using the link:https://ionicframework.com/docs/native/push[Ionic Push plugin].
+[source,javascript]
```
jhellar: I think that this line can be removed.

josemigallas: Good catch!

PR #45, opened 2018-06-12
Update steps for setting up iOS push notifications

jhellar:
# Motivation

https://issues.jboss.org/browse/AEROGEAR-3131

# Changes

- steps replaced with links to official documentation


wtrocki:
Technically this makes sense and it's provides all developers needs. 
It's also great from maintenance point of view. Leaving approval for docs technical committee and docs team. It will be nice to do the same for Android.

PR #46, opened 2018-06-13
MOBDOC-39 security service

finp:
@pb82  I'd suggest merging and then adding material


modules/ROOT/pages/push/bindin.adoc
```diff
@@ -0,0 +1,8 @@
+include::{partialsdir}/attributes.adoc[]
```
pb82: filename: `bindin` -> `binding`

PR #51, opened 2018-06-14
[PROPOSAL] Redesign "Prepare Development Environment" and "Adding code" section

josemigallas:
**Do not merge**

A redesign of current _Prepare Development Environment_ and _Adding code_ sections.

The logical flow on order to use a module from the AeroGear SDK is:

1. Provision service in Openshift
2. Bind application to the service
3. Get configuration file
4. Install SDK in mobile application
5. Initialize and use SDK

I think the current names do not describe these steps accurately and are even a bit confusing from the point of view of a mobile developer. My proposal is to have:

* Service
  * Provisioning
  * Binding
  * Configure your app
    * Getting config file
    * Adding SDK module
    * Custom section (ie Sending Metrics)

EDIT: custom section should be out of configuration. Following current Push approach:

* Service
  * Provisioning
  * Binding
  * Configure your app
    * Getting config file
    * Adding SDK module
  * More...
    * Custom section 1 (ie Sending Metrics)
    * ...


wtrocki:
@josemigallas If we do not comment that this PR will die. Proposal makes sense.

PR #54, opened 2018-06-15
'Configuring development environment' work introduced bugs

jstaffor:


jstaffor:
Tested locally and everything works as expected.

PR #56, opened 2018-06-15
Mobdoc 44 | Apply structure to all docs as a result of 'generic-obtaining-the-mobile-sdk-config-file' work

jstaffor:
@finp - can you take a look at this from the POV of the pattern being applied as you expect?


modules/ROOT/pages/keycloak/configuring-dev-env-keycloak.adoc
```diff
@@ -0,0 +1,18 @@
...
+:service-name: Identity Management
+
+= Configuring your Development Environment for the {service-name} Service
+
+== Prerequisites
+
+* [x] You have provisioned one of the {product-name} as described in xref:keycloak/provisioning.adoc[Provisioning {service-name} Service].
+* [x] (Optional) You have associated the service with a {mobile-client} as described in xref:keycloak/binding.adoc[Binding a {mobile-client} with the {service-name} Service].
+
+NOTE: You may skip creating a binding if only interested in metrics from other services but not from a {mobile-client}.
```
wtrocki: What this means? I do not understand that

jstaffor: So, this PR is all about a pattern being applied to the docs. A part of this work was to apply generic files across the different services - ie: get the structure in place. As evident by this question, some non generic info was brought with the generic files.

In summary, this looks like specific 'metrics' content - I'll pull it out of here unless you think the following would make sense:

"+NOTE: You may skip creating a binding if only interested in Identity Management from other services but not from a {mobile-client}."

wtrocki: @jstaffor I do not understand this sentence, but I'm fine anyway.

modules/ROOT/nav.adoc
```diff
@@ -17,40 +17,51 @@ include::{partialsdir}/attributes.adoc[]
...
-*** xref:configuring-dev-env.adoc[Configuring a Dev Environment]
+*** xref:push/configuring-dev-env-push.adoc[Configuring a Dev Environment]
 *** xref:push/coding.adoc[Adding Code]
 *** xref:push/admin-ui.adoc[{push-ui}]
 *** More...
 **** xref:push/proc_sending-a-push-notification.adoc[Sending a Push Notification]
 **** xref:push/handling-push-notifications.adoc[Handling Push Notifications]
 **** xref:push/proc_preparing-ups-for-use-with-ios.adoc[Apple Push Network]
 **** xref:push/google-setup.adoc[Firebase Cloud Messaging]
+
```
josemigallas: @jstaffor these blank lines will break the structure in the left hand menu, it has happened to me before.

modules/ROOT/pages/configuring-dev-env.adoc
```diff
@@ -7,196 +7,6 @@ include::{partialsdir}/attributes.adoc[]
 * [x] You have provisioned one of the {product-name} as described in xref:provisioning-services.adoc[Provisioning {product-name}].
 * [x] You have associated the service with a {mobile-client} as described in xref:binding-mobile-app-record-to-mobile-services.adoc[Binding a Mobile App Record with Mobile Services].
```
josemigallas: Could you remove the stars here? It should be just `[x]`

modules/ROOT/pages/ios/using-security-sdk.adoc
```diff
@@ -87,7 +87,7 @@ let results = security.checkMany(checks)
 ----
 
 == Publishing the Security Check Metrics
-This SDK provides functions which allow the developer to perform security checks and publish the results to the metrics server. To configure this, provide the metrics server configuration in the mobile-services.json file. If no 
+This SDK provides functions which allow the developer to perform security checks and publish the results to the metrics server. To configure this, provide the metrics server configuration in the mobile-services.json file. If no
 configuration is provided then the metrics publishing will fail silently and metrics will not be published.
```
josemigallas: I think this belongs to line 90.

modules/ROOT/nav.adoc
jstaffor: Thanks Jose - addressed in 857140205782ec7da1be2900b3635904ca5df4e0

PR #60, opened 2018-06-18
Create docs for 'User Authentication' - Android

jhellar:
# Motivation

https://issues.jboss.org/browse/AEROGEAR-3312

# Changes

Added document about User Authentication for Android.


modules/ROOT/pages/keycloak/android.adoc
```diff
@@ -0,0 +1,311 @@
+include::{partialsdir}/attributes.adoc[]
```
josemigallas: Instead of creating a new section called Android, use the same structure we have been using in Push or Metrics. Reuse the section "config-dev-env" which already include de info about config JSON and installing the SDK.
Then, under the "adding-code" section you add the specifics for Android in the tabbed box, see:
`mobile-docs/build/site/aerogear/latest/push/coding.html`

PR #66, opened 2018-06-18
MOBDOC-45 change release-number to installer-release-number

pmacko1:
https://issues.jboss.org/browse/MOBDOC-45

### DIFF
--- a/modules/ROOT/pages/_partials/attributes.adoc
+++ b/modules/ROOT/pages/_partials/attributes.adoc
@@ -2,7 +2,7 @@
 
 :product-name: Mobile Services
 
-:release-number: 1.0.0-alpha
+:installer-release-number: 1.0.0-alpha
 
 :service-name:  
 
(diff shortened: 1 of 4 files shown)


pb82:
@finp No worries, i can rebase my PR once you have merged this. But why are we renaming it to `installer-release-number` when it's used for other artifacts too (SDKs and APBs are also tagged with that version)?

finp:
@pb82  can you confirm that all software will adhere to that version in future. My understanding is that release versions for these will all happen async

pb82:
@finp According to @maleck13 this hasn't been decided yet.

finp:
@jasonmadigan  We need to decide this issue, currently we're using {release-no} as if we're releasing a platform, not a set of services/sdks/other modules that can be versioned independently

finp:
after discussion with @johnfriz  closing this until after community release

jasonmadigan:
@finp grand 👍

PR #74, opened 2018-06-19
DO-NOT-MERGE: Changes proposal to structure and contents

josemigallas:
### Proposed changes
- GETTING STARTED contains a list of steps that can be read 1 after another (chained by “next steps” links) and by the last step, the user should have their environment ready.
- Each service in USING MOBILE SERVICES, contains a list of steps that are needed to provision, configure and initialize it, and after those a list of custom actions that describe the functionality of the service’s SDK (ie for Push: Registering a Device, Handling Push Notifications, etc).
- CONFIGURING A DEV ENVIRONMENT doesn’t contain generic steps about installing the AeroGear SDK but only the steps and code snippets required for the particular service. A generic description with code examples would be located under GETTING STARTED > CONFIGURING A DEV ENVIRONMENT
- Remove USING MOBILE SERVICES and bring all services up 1 level.

### Other thoughts
- I feel confusing that each title in the left hand menu can be selected and has content, but it’s different than its subsections. For clicking on GETTING STARTED will show an “Introduction to Mobile Services” document but the first content should be only INSTALLING. I think GETTING STARTED could be a big document where its subsections INSTALLING, TERMINOLOGY, etc are parts of it. This way, even when clicking on INSTALLING for the first time (instead of GETTING STARTED) the user would be able to see that there is some content before.


finp:
closing in favor of #119

PR #76, opened 2018-06-19
update build scripts and readme

finp:


README.adoc
```diff
@@ -133,6 +133,9 @@ Currently, the publishing process is manual:
 ----
 
 . Review the output (`build\site\index.html`) for errors. Correct if required.
++
+NOTE: Check for correct styling changes too.
```
finp: +1
I think, over time, we'll add to this, ie a list of checks

pb82:
With this change it's best to use the antora cli directly to build the local site: `antora --clean --pull local-site.yml`

finp:
@pb82  running antora, rather than quick-build.sh, won't render the docs correctly, not sure what your comment relates to?

pb82:
@finp build.sh used to build `local-site.yml` and i was using that to test local changes. But looks like i can use `quick-build.sh` too.

PR #78, opened 2018-06-19
AEROGEAR-3349 Created the structure for showcase apps

aliok:
Created the structure for showcase apps. No content, just the structure.

https://issues.jboss.org/browse/AEROGEAR-3349


aliok:
This is outdated as we have a new freezed structure now.

PR #81, opened 2018-06-19
Aerogear 3355 intro to showcase apps

aliok:
Depends on https://github.com/aerogear/mobile-docs/pull/78


modules/ROOT/pages/showcase-apps/index.adoc
```diff
@@ -0,0 +1,8 @@
+include::{partialsdir}/attributes.adoc[]
+
+= Showcase Apps
+
+There are showcase apps available in all mobile supported platforms that show the provided functionality.
```
pb82: We usually link the content here from `service-intros.adoc` but i'm not sure if it makes sense for the showcase apps.

aliok: nah, I wasn't able to make that work :( Would look ugly.

PR #82, opened 2018-06-19
Aerogear 3312

pb82:
Cleanup and refactor the IDM docs to better adhere our docs structure. Also added code snippets for all platforms. This is what the menu now looks like (all sections have content):

bildschirmfoto 2018-06-19 um 22 01 47


aliok:
@pb82 I don't know if it was intentional, but I don't see anything related to getting user attributes. For example: https://github.com/aliok/mobile-docs/blob/a48f433aac7a784053a9d547e2c8c03e7c04ab60/modules/ROOT/pages/keycloak/retrieving-current-user.adoc

Also similarly, no getting custom attrs: https://github.com/aliok/mobile-docs/blob/a48f433aac7a784053a9d547e2c8c03e7c04ab60/modules/ROOT/pages/keycloak/retrieving-custom-user-attributes.adoc

modules/ROOT/pages/keycloak/retrieving-users-roles.adoc
```diff
@@ -1,52 +1,103 @@
...
 ****
 
 [role="secondary"]
 .Cordova
 ****
+Get the realm roles from the auth service:
+
+[source,javascript]
+----
+const realmRoles = authService.getRealmRoles();
```
aliok: In Cordova, we miss some stuff. Created https://issues.jboss.org/browse/AEROGEAR-3350 for that.

But, we at least have `hasRealmRole` method.
I guess you didn't want to have that here?
https://github.com/aliok/mobile-docs/blob/a48f433aac7a784053a9d547e2c8c03e7c04ab60/modules/ROOT/pages/keycloak/retrieving-users-roles.adoc

pb82:
@aliok Skipping the user attributes and custom user attributes was intentional. For the normal user attributes i'm not sure if we really need to document them. They are kind of obvious and (as long as the user has code completion) can be discovered easily.

Regarding the custom user attributes: i can add that section back in (it's still in the repo) but it's extra content that is not required to use the auth service.

modules/ROOT/pages/keycloak/retrieving-users-roles.adoc
pb82: Yes, we miss some function in the js-sdk. I've covered this in the following note and suggest to use keycloak-js directly:

```
NOTE: No function to retrieve the resource roles is currently exported but you can use `extract` to get a handle to the underlying keycloak-js instance to retrieve the roles from there. Have a look at their documentation link:https://www.keycloak.org/docs/latest/securing_apps/index.html#_javascript_adapter[here].
```

modules/ROOT/nav.adoc
```diff
@@ -39,14 +39,11 @@ include::{partialsdir}/attributes.adoc[]
 *** xref:keycloak/binding.adoc[Binding]
 *** xref:keycloak/configuring-dev-env-keycloak.adoc[Configuring a Dev environment]
 *** xref:keycloak/coding.adoc[Adding Code]
+*** xref:keycloak/authenticating.adoc[User Authentication]
```
josemigallas: Should this be called `Authenticating a User` instead? @finp

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -34,13 +36,8 @@ Callback authCallback = new Callback<UserPrincipal>() {
...
 ----
 
-Once the browser returns to the app the result can be handled. In the activity provided to the
-`login` method override `onActivityResult`. This handler should always invoke
-`AuthService#handleAuthResponse`, providing the `Intent`. This will exchange the temporary tokens
-returned from `AuthService#login` for long-life tokens and will provide a `UserPrincipal` which can
-be used to access a users details. If this is not invoked you will not have access to the
-`UserPrincipal`.
-
+. Override `handleAuthResult` to handle the result from the browser
```
josemigallas: This should be:
In `onActivityResult`, pass the Intent to your instance of `AuthService`.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -34,13 +36,8 @@ Callback authCallback = new Callback<UserPrincipal>() {
...
++
 [source,java]
 ----
 @Override
```
josemigallas: A different code is being used to get the instance here.
`mobileCore.getInstance(AuthService.class)`
And in the previous one is
`MobileCore.getInstance().getService(AuthService.class)`
This is confusing, the former one is an instance of MobileCore but the latter is calling a static method.
Let's be consistent. I think this has to work as is so I think the correct one is the static method.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
         authService.handleAuthResult(data);
     }
 }
+----
++
+NOTE: This handler should always invoke `handleAuthResult`, providing the `Intent`. This will exchange the temporary tokens returned from `login` for long-lived tokens and will provide a `UserPrincipal` which can be used to access a users details. If this is not invoked you will not have access to the `UserPrincipal`.
```
josemigallas: I think this note is not necessary, it is what the 2nd step describes anyway you have to pass the intent to `handleAuthResult`.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+
+. The callback provided to `login` will be invoked.
+****
+
+[role="secondary"]
+.iOS
+****
+. Use the `login` function to trigger the authentication flow:
```
josemigallas: Let's be consistent and use the same wording for each platform, if possible.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
++
+[source,swift]
+----
+try AgsAuth.instance.login(presentingViewController: self, onCompleted: onLoginComplete)
+
+// handle the authentication result
+func onLoginComplete(user: User?, err: Error?) {
+    // login failed
```
josemigallas: Apologies for putting the blame on your PR, but I won't rest until our code base is cleansed of redundant comments. The code is clear enough, and I do think it has a negative impact on future readers.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+    }
+    return false
+}
+----
+****
+
+[role="secondary"]
+.Cordova
+****
+Authentication will be performed on initialization as described xref:keycloak/coding.adoc[here]. To manually redirect to the login screen use:
```
josemigallas: Change URL to point to correct section `file:///Users/jgallaso/Projects/aerogear/mobile-docs/build/site/aerogear/latest/keycloak/coding.html#initializing-the-sdk`

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+[code,javascript]
+----
+authService.login()
```
josemigallas: I think we should implement the `then` method too.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+----
+****
+
+[role="secondary"]
+.Xamarin
+****
+. Get an instance of `IAuthService`:
```
josemigallas: Again let's be consistent. This step of _getting an instance_ is not explicit in the other platforms. I am in favor of having this step, but then let's add it to Android, iOS and Cordova.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+----
+
+. Use the `Authenticate` method to start the auth flow:
++
+[source,csharp]
+----
+var authOptions = DependencyService.Get<IAuthenticateOptionsProvider>().GetOptions();
+service.Authenticate(authOptions).ContinueWith(result =>
+{
+    // load some view
```
josemigallas: I think we should do something with "result" here. What's result? Is it the user already? Is it some login result info? Could it be an error message?

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+});
+----
+****
+
+== Logging out
```
josemigallas: Why log out has a title but log in doesn't?

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+
+The following section describes how to perform a logout.
+
+[role="primary"]
+.Android
+****
+To logout, invoke the `logout` method. This accepts the UserPrincipal that was provided by `handleAuthResult` and has a callback to determine if the logout to the Keycloak or OpenID Connect server was successful.
```
josemigallas: Let's be consistent and describe this procedure in the same fashion as the login. 1. Get instance.. 2.implement callback and call logout...

There the callback was a separate variable while here it's anonymous.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+
 ----
 
-The callback provided in `AuthService#login` will be invoked.
 ****
 
 [role="secondary"]
 .iOS
 ****
+. To logout invoke the `logout` method:
```
josemigallas: I would change the order, first implement callback then call logout method. Also I think we could reword _post-logout actions_ for simply results or _check for errors_

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-****
\ No newline at end of file
+To logout invoke the `logout` function:
+
+[source,javascript]
+----
+authService.logout()
+----
+
+By default, the local tokens obtained during authentication are only deleted when the logout succeeded against the authentication server.
```
josemigallas: What about adding a NOTE for this?

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+
+****
+
+[role="secondary"]
+.Xamarin
+****
+To logout invoke the `Logout` method:
+[source,csharp]
+----
+// auth service already initialised
```
josemigallas: This does not make much sense.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+
+var authService = MobileCore.Instance.GetInstance<IAuthService>();
+authService.Logout(authService.CurrentUser()).ContinueWith(result =>
+{
+    // post-logout actions
```
josemigallas: Same comments as in Xamarin's login snippet

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -51,19 +48,180 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+});
+----
+****
+
+NOTE: To perform backchannel or federated logouts, you must enable the Backchannel Logout option for the federated identity provider. More information is available in the Keycloak documentation under OIDC Identity Providers.
```
josemigallas: Let's add a link for that documentation and can we headlight `Backchannel Logout` option somehow (bold)?

modules/ROOT/pages/keycloak/binding.adoc
```diff
@@ -5,3 +5,7 @@ include::{partialsdir}/attributes.adoc[]
 = Binding a {mobile-client} with the {service-name} Service
 
 include::{partialsdir}/generic-binding.adoc[]
+
+. You need to choose the *Keycloak client type*. The default is `public` which allows the client to request tokens while type `bearer` should be used to bind to services that only need to verify tokens.
```
josemigallas: I think the preferred way is imperative tense `Choose the Keycloak client type` instead of `You need to...`.

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -9,74 +9,123 @@ This guide will help you to set up the {keycloak-service} service SDK in your Ap
 * xref:keycloak/provisioning.adoc[Provisioned {keycloak-service} service].
 * xref:keycloak/binding.adoc[Binding] created.
 * xref:keycloak/configuring-dev-env-keycloak.adoc[Configured dev environment].
-* xref:keycloak/keycloak-setup.adoc[Configured Keycloak].
+* Read the xref:configuring-dev-env.adoc[general app setup guide].
```
josemigallas: Can we have this prerequisites in the same fashion that "Configuring a Dev environment"?

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -9,74 +9,123 @@ This guide will help you to set up the {keycloak-service} service SDK in your Ap
...
+    pod 'AGSAuth', '{release-number}'
+end
+----
+
+. Update the dependencies
 +
+[source,bash]
 ----
-pod 'AGSAuth'
+pod install
```
josemigallas: missing $ here, see https://github.com/aerogear/mobile-docs/blame/master/modules/ROOT/pages/push/coding.adoc#L72

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -9,74 +9,123 @@ This guide will help you to set up the {keycloak-service} service SDK in your Ap
...
 ----
 ****
 
 [role="secondary"]
 .Cordova
 ****
+Install the auth package:
+[source,bash]
+----
+npm install @aerogear/auth
```
josemigallas: Missing $

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -9,74 +9,123 @@ This guide will help you to set up the {keycloak-service} service SDK in your Ap
...
-// Specify redirect URI. It is recommended to use the package name of your app.
 AuthServiceConfiguration authServiceConfig = new AuthServiceConfiguration
     .AuthConfigurationBuilder()
     .withRedirectUri("org.aerogear.mobile.example:/callback")
     .build();
+----
 
-// You only need to invoke this once every subsequent retrieval of the AuthService
-// will retrieve the same instance.
+. Before the AuthService can be used `init` must be invoked once in an app
```
josemigallas: In step 1 it refers to `the auth service` but here is in Pascal case, like a class `the AuthService`. If sticking with class notation, highlight using ``.

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -9,74 +9,123 @@ This guide will help you to set up the {keycloak-service} service SDK in your Ap
...
++
+[source,java]
+----
 authService.init(context, authServiceConfig);
 ----
 
-Follow xref:keycloak/android-additional-configuration.adoc[this guide] for additional configuration.
+. Follow xref:keycloak/android-additional-configuration.adoc[this guide] for additional configuration.
```
josemigallas: Let's say explicitly what is to be followed: For additional configuration see [Additional configuration for Android](https://www.github.com/aerogear/mobile-docs/).

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -9,74 +9,123 @@ This guide will help you to set up the {keycloak-service} service SDK in your Ap
...
 ****
 
 [role="secondary"]
 .iOS
 ****
-The Auth SDK can be initialised with configuration options. Use the redirect URL xref:keycloak/keycloak-setup.adoc[configured in Keycloak].
+The Auth SDK can be initialised with configuration options. The redirect URL should match the App’s Bundle Identifier.
```
josemigallas: Let's be imperative, something like _Set your custom configuration to the auth service instance, making sure the redirect URL matches the App's Bundle Id_.
And let's remove redundant comments from code examples.

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -90,4 +139,57 @@ try! AgsAuth.instance.configure(authConfig: authenticationConfig)
 .Cordova
 ****
 
-****
\ No newline at end of file
+Import and initialize Auth:
```
josemigallas: _... initialize Auth_ is what the section is to describe (it's called initializing the sdk), what about _Instantiate `Auth` and call init method_?

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -90,4 +139,57 @@ try! AgsAuth.instance.configure(authConfig: authenticationConfig)
...
+[source,csharp]
+----
+var app = new App();
+MobileCoreAndroid.Init(app.GetType().Assembly,ApplicationContext);
+var authService = AuthService.InitializeService();
+var authConfig = AuthenticationConfig.Builder.RedirectUri("org.aerogear.mobile.example:/callback").Build();
+authService.Configure(authConfig);
+----
+
+NOTE: For Android an link:https://developer.android.com/guide/topics/manifest/intent-filter-element[intent-filter] should be configured with the callback URL specified in AuthenticateOptions in the apps AndroidManifest.xml. See the link:https://github.com/aerogear/xamarin-showcase-template/blob/master/Example.Android/Properties/AndroidManifest.xml[example app].
```
josemigallas: - `Intent filter` is more appropriate, also link to https://developer.android.com/guide/topics/manifest/manifest-intro#ifs
- I think `apps AndroidManifest.xml` -> `app's AndroidManifest.xml`.
- Let's add this intent filter requirement as a step instead of pointing to the example app, if it's necessary.

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -90,4 +139,57 @@ try! AgsAuth.instance.configure(authConfig: authenticationConfig)
...
+You can pass `login-required` or `check-sso` to the init function. `login-required` will authenticate the client if the user is logged in to Keycloak or display the login page if not. `check-sso` will only authenticate the client if the user is already logged in. If the user is not logged in the browser will be redirected back to the application and remain unauthenticated. By default, the `check-sso` option is used.
+
+NOTE: Initialization will also perform authentication
+
+****
+
+[role="secondary"]
+.Xamarin
+****
+*Initializing Auth in an Android app (MainActivity.cs)*
```
josemigallas: We could reword a bit:
_For initializing Auth in Android, add the following code to your `MainActivity.cs`._
_For initializing Auth in iOS, add the following code to your `ApDelegate.cs`._

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -1,12 +1,14 @@
 include::{partialsdir}/attributes.adoc[]
 
-= Authenticating
+= User authentication
```
josemigallas: This sections comes directly from "Adding Core", therefore I think we should add a "prerequisites" paragraph first with
[x] - You have imported the necessary libraries and initialized the SDK

modules/ROOT/pages/keycloak/authenticating.adoc
pb82: I'll change it to the correct `onActivityResult` but i think we should not describe the intent passing here because it's explained in a note afterwards. Will also remove the comment from the snippet.

modules/ROOT/pages/keycloak/authenticating.adoc
pb82: Yes, as much as possible. But for example in Java you would call it 'Method' while Javascript doesn't really have the concept of a method and you would just say 'function'. Or in Java it would be somewhat verbose to not retrieve the AuthService instance first and just chain the calls while in Swift that's the nice and concise way. So sometimes the procedure will be slightly different.

modules/ROOT/pages/keycloak/authenticating.adoc
ziccardi: Authenticate returns a `Task<User>`. ContinueWith is executed when the task is executed and result is the `finished `Task<User>`. Look https://msdn.microsoft.com/library/dd270696(v=vs.110).aspx

modules/ROOT/pages/keycloak/authenticating.adoc
pb82: Login is basically `User authentication`

modules/ROOT/pages/keycloak/authenticating.adoc
ziccardi: As discussed on IRC, if the enclosing method is marked with `async` this could be changed to:
```csharp
var user = await service.Authenticate(authOptions);
// load some view.
```

pb82: @ziccardi Much nicer, thanks. For view switching i'm using: `Device.BeginInvokeOnMainThread(() => ((RootPage)(App.Current.MainPage)).ChangePage(new UserDetails()));`

modules/ROOT/pages/keycloak/authenticating.adoc
pb82: It kind of does for Xamarin because it requires you to initialize the core in a previous step. I have added docs for that to the general dev env setup guide. But i agree that the comment is confusing and i will remove it.

modules/ROOT/pages/keycloak/authenticating.adoc
pb82: The link can be found in `Additional resources` but i add it here too.

modules/ROOT/pages/keycloak/coding.adoc
pb82: I'm not sure. If people copy the whole line to paste into a terminal they dont want the $. But we should be consistent.

modules/ROOT/pages/keycloak/coding.adoc
pb82: This document will no longer be included in the 1.0.0-alpha release

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -90,4 +139,57 @@ try! AgsAuth.instance.configure(authConfig: authenticationConfig)
 .Cordova
 ****
 
-****
\ No newline at end of file
+Import and initialize Auth:
+
+[source,javascript]
+----
+const authService = new Auth();
```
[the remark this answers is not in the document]
pb82: We can't really assume which (if any) module loader the user is using. require? import? But i think it's ok to make an example using require.

modules/ROOT/pages/keycloak/coding.adoc
ziccardi: @pb82 : not sure if this can impact, but for Xamarin we have already 1.0.0-alpha-1 and we are going to have 1.0.0-alpha-2 soon (we just made important fixes into the SDK that should be released to be used by the showcase)

modules/ROOT/pages/keycloak/coding.adoc
```diff
@@ -101,5 +139,76 @@ try! AgsAuth.instance.configure(authConfig: authenticationConfig)
...
+
+NOTE: Initialization will also perform authentication
+
+****
+
+[role="secondary"]
+.Xamarin
+****
+
+. Create an link:https://developer.android.com/guide/topics/manifest/manifest-intro#ifs[intent filter] for the `net.openid.appauth.RedirectUriReceiverActivity` activity. This step is required for Xamarin Android and allows the login browser to redirect back to your App. Add this to your `AndroidManifest.xml`:
```
pb82: @secondsun does this make sense to you?

secondsun: +1

modules/ROOT/pages/keycloak/authenticating.adoc
finp: The procedure should describe what the user is trying to achieve,
Would 'Adding User Authentication to your Mobile App' make sense?

modules/ROOT/pages/keycloak/authenticating.adoc
finp: How about
Adding Log Out Option to your Mobile App

modules/ROOT/pages/keycloak/coding.adoc
finp: @pb82 consistency is the most important, but also checked with other  writers, let's go with $
I know it's not ideal, but we can change later if required (easy to change if we're consistent)

PR #101, opened 2018-06-22
update showcase apps docs and self-signed cert section

finp:


modules/ROOT/pages/showcase-apps/service-setup.adoc
```diff
@@ -1,24 +1,28 @@
...
 . Bind your {mobile-client} to the appropriate services as described in xref:binding[]
-. Copy the `mobile-services.json` file to your IDE.
+. Copy the `mobile-services.json` file to your IDE as described in xref:downloading-the-mobile-services-configuration-file[]
 . Build and run your showcase app  as described in xref:building-and-deploying-the-showcase-apps[].
 
 [[setting-up-mobile-services-to-demonstrate-showcase-apps]]
 == Setting up Mobile Services to Demonstrate Showcase Apps
 
-
+[discrete]
```
pb82: What does `discrete` do?

finp: shared answer in other channel
 [discrete] suppresses a heading so that it doesn't appear in a toc
one of things I do when editing an adoc is add :toc: temporarily to the start of a doc, then I can see if the flow of the document works while I'm editing
[discrete] stops things like == Next Steps from cluttering that view

PR #107, opened 2018-06-25
Ups fixes

pb82:
some smaller fixes for the UPS section + removal of some details that dont need to be in the docs.

### DIFF
--- a/modules/ROOT/pages/push/index.adoc
+++ b/modules/ROOT/pages/push/index.adoc
@@ -19,25 +19,9 @@ Currently the service supports:
 * link:https://developer.apple.com/notifications/[Apple Push Notification Service]
 * link:https://firebase.google.com/products/cloud-messaging/[Firebase Cloud Messaging]
 
-== Use-cases and scenarios
-
-Different use-cases and scenarios are supported. Below are a few to give an idea how the UnifiedPush Server can be used:
-
-* MyWarehouseInc-backend can send _notification messages_ to different groups (e.g. discounts for only iOS (or only Android) users)
-* MyInsuranceCorp-backend can send “notification messages” to different variants of its mobile Applications:
-  ** App for the Customers
-  ** App for the employed Sales Agents
-* Publishing Company:
-  ** MyPublishing-Company-backend sends update “notification messages” to all of its apps (free and premium - regardless of the mobile OS)
-  ** Targeting: Sending push messages to different groups of users. For instance, availability of “advanced content” is only notified to the paying customers (e.g. those that run the premium app).
-* A company has different backends (small/simple apps for different tasks) - and these different backends could be able to reach all (or some) of the company’s mobile apps.
-
-**__Motivation:__** Easy use of Push Notifications for any (JBoss/AeroGear powered) mobile application, that is backed by JBoss technology (e.g. LiveOak, Drools, etc.).
-
 include::./ref_terminology.adoc[leveloffset=+1]
 
-
-=== Privacy note
+== Privacy note
 
 As explained in the _link:#_how_the_unifiedpush_server_works[How the UnifiedPush Server works]_ section, the payload of the push notification is delivered to 3rd party Push Network providers, like Google or Apple.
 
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/push/index.adoc
```diff
@@ -19,25 +19,9 @@ Currently the service supports:
 * link:https://developer.apple.com/notifications/[Apple Push Notification Service]
 * link:https://firebase.google.com/products/cloud-messaging/[Firebase Cloud Messaging]
 
-== Use-cases and scenarios
```
cfoskin: Are we just not providing any use cases?

modules/ROOT/pages/_partials/registering-a-mobile-app.adoc
```diff
@@ -1,6 +1,6 @@
 include::{partialsdir}/attributes.adoc[]
 
-= Registering a Mobile App
+= Registering a {mobile-client}
```
finp: you register a mobile app (to create a mobile-client)
ie there's no mobile-client before you register, and the only thing that exists (even if it's just in your head) is a mobile app

modules/ROOT/pages/push/provisioning.adoc
```diff
@@ -1,6 +1,6 @@
 include::{partialsdir}/attributes.adoc[]
 
-:service-name: Push Notifications
+:service-name: Unified Push Server
```
finp: what's this? should be same as attributes.adoc

modules/ROOT/pages/push/provisioning.adoc
```diff
@@ -14,4 +14,6 @@ include::{partialsdir}/generic-provisioning-end.adoc[]
 
 == Next Steps
 
+xref:registering-a-mobile-app.adoc[Registering a {mobile-client}]
```
finp: as above

modules/ROOT/pages/push/provisioning.adoc
pb82: I changed it to `Unified Push Server` because thats the name you see in the service catalog. It would be confusing for the provisioning instructions to call it `Push Notification Service` when the name of the APB actually is `Unified Push Server`.

Or should we change the APB name / change the service name in attributes.adoc?

modules/ROOT/pages/push/index.adoc
```diff
@@ -46,21 +30,13 @@ WARNING: It is highly recommended to not sent any send sensitive personal or con
 For analytic purposes on our link:#_dashboard[Dashboard] we store the content of the +alert+ key sent to the UnifiedPush Server. The content of the +alert+ key belongs to the metadata, which is deleted after 30 days, using a nightly job within the UnifiedPush Server.
 
 
-=== How the UnifiedPush Server Works
+== How the UnifiedPush Server Works
 
-The AeroGear UnifiedPush Server offers a unified Notification Service API to the above mentioned Push Network Services. When a push message request is sent to the UnifiedPush Server, it is internally translated into the format of these 3rd party networks. This gives a server the ability to send Push notifications to different mobile platforms.
+The AeroGear UnifiedPush Server offers a unified Notification Service API to the above mentioned Push Network Services.
```
josemigallas: The AeroGear "UnifiedPush Server" should be "{unifiedpush-service}" here?

finp: yep

modules/ROOT/pages/push/provisioning.adoc
josemigallas: Peter you have a point here. IMO the confusion comes with the name `Unified Push Server`, it's inaccurate and I tend to associate it to the UI only. It would make more sense to find it as **Push Notifications** or simply **Push** in the catalog. Then, I propose to leave it here as `Push Notifications` for now and try to change APB's name.

finp: @pb82  Name needs to be changed in service catalog -ping @maleck13

PR #108, opened 2018-06-25
prefer UI approach for NuGet packages

pb82:
Prefer the VS UI approach to install the Xamarin core packages.

### DIFF
--- a/modules/ROOT/pages/configuring-dev-env.adoc
+++ b/modules/ROOT/pages/configuring-dev-env.adoc
@@ -130,62 +130,18 @@ init(aerogearConfig);
 
 ****
 
-. Install link:https://docs.microsoft.com/en-us/nuget/install-nuget-client-tools[NuGet]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/configuring-dev-env.adoc
```diff
@@ -136,11 +136,49 @@ init(aerogearConfig);
...
+[souce,csharp]
+----
+MobileCoreAndroid.Init(ApplicationContext);
+----
+
+. for iOS:
+
+.. also pick `AeroGear.Mobile.Core.Platform.iOS`
+
+.. Click on `Add Packages`
```
ziccardi: I believe this shoulbe before the previous bullet

pb82: Actually this refers to the `Add Packages` button in the dialog where you add the package. But it looks a bit confusing. I think i just remove the whole bullet point.

PR #119, opened 2018-06-27
Restructure

finp:


finp:
1. Restructure nav.adoc
2. refactor getting started into one doc
3. make all service docs a single doc

PR #124, opened 2018-06-29
fix link formatting

laurafitzgerald:
**What**
Fix on linking in a number of locations in this file: https://finp.github.io/aerogear/latest/mobile-cicd.html

**Before**
image

**After**
image

@finp


johnfriz:
👍

PR #128, opened 2018-07-01
update readme for running antora with docker

finp:


finp:
@dimitraz @pawelpaszki  Could you take a look at this (running antora using docker, rough draft)?

finp:
didn't work on Fedora 28

PR #137, opened 2018-07-02
add a screenshot for registering the mobile clients in getting starte…

wei-lee:
…d guide

JIRA: https://issues.jboss.org/browse/AEROGEAR-3510

Add a screenshot of the mobile client view when registering a mobile client.

ping @finp

### DIFF
--- a/modules/ROOT/pages/_partials/registering-a-mobile-app.adoc
+++ b/modules/ROOT/pages/_partials/registering-a-mobile-app.adoc
@@ -21,4 +21,4 @@ You can filter the catalog items to only show mobile specific items by clicking 
 image:mobile-platforms.png[Available Mobile Clients]
 . Click the *Apps* tab and then choose the mobile platform (Android, iOS, Cordova or Xamarin) and follow the wizard.
 
-After the {mobile-client} is provisioned, you can navigate to it from Project Overview. This screen displays a list of mobile services that you can associate with the {mobile-client} and offers to provision any mobile service that is in the service catalog but is not currently provisioned. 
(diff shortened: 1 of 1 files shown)


johnfriz:
@wei-lee Can you add a screenshot of how the docs look with this image in place. Will make reviewing much quicker.

wei-lee:
screen shot 2018-07-02 at 13 14 36

wei-lee:
@johnfriz @finp updated now. I can also move it be part of step 4 if you think it's better that way.

johnfriz:
Looks good overall.

Not sure about the last sentence though "This screen displays a list of mobile services
that you can associate with the Mobile Client and offers to provision any mobile service that is in the service catalog but is not currently provisioned."

What does the "this" refer to at the start of the sentence - is it the project overview screen? If so, then I'm not sure the content is correct. If it's not the overview screen, then we need to clarify what "this" means.

johnfriz:
@wei-lee - OK, looking at the source of the changes, I see that you did not write/update that sentence. However, I still think we need to change it - should it be a separate ticket or just done in this PR?

johnfriz:
@wei-lee - It also still bothers me that the Android and iOS icons are still black & white while the Cordova and Xamarin ones are in colour. I know it's not related to this PR, but wondering if we want to fix it for community release? If so, we would need to re-do this screen shot. Thoughts?

wei-lee:
@johnfriz I update the wording to replace "This" with "The {mobile-client} view". I think once we have the mobile client section done, we can link to it.

In terms of the icons, they are still white & black in the latest version of the web console. I don't know how much work there will be to update them, I think @jasonmadigan or @sedroche will have a better idea. I don't mind updating the screenshot again if they are updated.

But for now, I'd like to merge this PR and when we have the icons we can update it again.

wei-lee:
I created a JIRA to capture the work to update the icons.

johnfriz:
👍

PR #153, opened 2018-07-02
Update the instructions about enable self-signed certificates

wei-lee:
screen shot 2018-07-02 at 18 19 57
screen shot 2018-07-02 at 18 19 49

ping @johnfriz @finp

### DIFF
--- a/modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
+++ b/modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
@@ -1,24 +1,23 @@
 //[[using-self-signed-certificates-in-mobile-clients]]
 = Using Self-Signed Certificates in Mobile Apps
 
-Throughout the development lifecycle of a mobile apps, a common
+Throughout the development lifecycle of a mobile app, a common
 requirement is to integrate and connect with back-end
-services in a secure manner. However in order for the client
-device to connect over SSL/TLS, it needs to trust the root
-certificate authority. This cert must be installed on the client device
-to ensure that the client trusts server certificates that are signed by
-the certificate authority.
+services in a secure manner. This is achieved using SSL/TLS. 
 
-When the certificate uses a trusted root certificate authority the cert is installed automatically.
-In a development environment, such as a local OpenShift cluster which uses
-self-signed certificates, you must extract the root certificate from the cluster, install it on the device and mark it as trusted.
+However, in order for the client device to connect over SSL/TLS, it needs to trust the
+certificates used by the back-end services, which are signed by a certificate authority. Today most of the client devices will have a list of well-known and trusted certificate authorities pre-installed and it will allow the client devices to verify the certificates used by the back-end services.
+
+However, this normally doesn't work in a development environment, such as a local OpenShift cluster as it uses self-signed certificates which are not signed by any of the trusted certificate authorties. This means by default the client devices won't be able to establish secure connections with the back-end services that are running on a local OpenShift cluster.
+
+To work around the problem, you must manually extract the root certificate from the cluster, install it on the device and mark it as trusted.
 
 [[extracting-root-ca]]
-== Extracting the OpenShift Root Certificate Authority Cert
+== 1. Extracting the OpenShift Root Certificate Authority Cert
 
-. Log into OpenShift as the admin user.
+.. Log into OpenShift as the admin user.
 
-. Run the following command:
+.. Run the following command:
 +
 [source,bash]
 ----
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
```diff
@@ -29,26 +28,25 @@ base64 --decode | sed -e '1,/^-----END RSA PRIVATE KEY-----$/ d'  > /tmp/localcl
...
 [[installing-on-device]]
-== Installing the OpenShift Root Certificate Authority Cert on a Device
+== 2. Installing the OpenShift Root Certificate Authority Cert on a Device
+
+[role="primary"]
+.Android
 
-. Configure a pin to ensure the certificate can be installed.
+****
+.. Enable PIN protection on the mobile device to ensure the certificate can be installed.
```
johnfriz: Is there a link to a android doc we could add here which gives instructions on how to do this?

modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
```diff
@@ -1,24 +1,23 @@
...
-In a development environment, such as a local OpenShift cluster which uses
-self-signed certificates, you must extract the root certificate from the cluster, install it on the device and mark it as trusted.
+However, in order for the client device to connect over SSL/TLS, it needs to trust the
+certificates used by the back-end services, which are signed by a certificate authority. Today most of the client devices will have a list of well-known and trusted certificate authorities pre-installed and it will allow the client devices to verify the certificates used by the back-end services.
+
+However, this normally doesn't work in a development environment, such as a local OpenShift cluster as it uses self-signed certificates which are not signed by any of the trusted certificate authorties. This means by default the client devices won't be able to establish secure connections with the back-end services that are running on a local OpenShift cluster.
+
+To work around the problem, you must manually extract the root certificate from the cluster, install it on the device and mark it as trusted.
 
 [[extracting-root-ca]]
```
johnfriz: I am wondering if this is the easiest way to get the cert trusted. How many devices would support the approach of just opening the URL of the local OpenShift cluster from the device browser and following the instructions presented by the browser to add the certificate. Pretty sure this works on Android, but not sure about the other platforms.

Thoughts?

modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
```diff
@@ -29,26 +28,25 @@ base64 --decode | sed -e '1,/^-----END RSA PRIVATE KEY-----$/ d'  > /tmp/localcl
...
 
-. Configure a pin to ensure the certificate can be installed.
+****
+.. Enable PIN protection on the mobile device to ensure the certificate can be installed.
 
-. Copy the file to your device using one of the following methods:
+.. Copy the file to your device using one of the following methods:
 +
-* USB *(Android only)* - Attach the device to the machine hosting
+* USB - Attach the device to the machine hosting
```
johnfriz: Do we want to a link to android documents for how to copy a file to Android using USB? The current instructions tell you what to do, but not how to do them.

modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
```diff
@@ -29,26 +28,25 @@ base64 --decode | sed -e '1,/^-----END RSA PRIVATE KEY-----$/ d'  > /tmp/localcl
...
-
-. Configure your device:
-
-[role="primary"]
-.Android
-
-****
+.. Configure your device:
++
 Go to Settings > Security > Install from storage. It should detect the
```
johnfriz: Again, is there a doc in Android docs that we can link to which explains this in more detail?

modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
```diff
@@ -57,11 +55,22 @@ certificate and let you add it to the device.
 .iOS
 
 ****
+.. Enable Passcode or TouchID protection on the mobile device to ensure the certificate can be installed.
```
johnfriz: Same as above - is there a link to iOS docs that we could include which takes a user through this

modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
wei-lee: I don't think there are a lot of devices support this approach. I tried it on my own android phone and even if I trusted the web console in the browser, it still doesn't work for me in the app.  Also from Chrome, I don't see there is an option to allow to download the CA file and save it on the device. Does it work on your phone?

modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
wei-lee: We can, but it could be different depending on the manufacture. So here I will link to the guide for Google Nexus and use it as an example.

wei-lee:
@johnfriz All the other comments are addressed. Mind taking another look? Thanks.

modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
johnfriz: OK. Let's leave it as is for now so.

johnfriz:
@wei-lee - Trying to follow these instructions and the cert can not be imported on the emulator. I get a message saying "No Certificate to Install".

The output from running the oc command to get the cert produces a file with 
-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
format, but the emulator (Nexus 5X API 26) does not seem to want to recognise it. 

Any thoughts ?

wei-lee:
@johnfriz I followed the steps as well and it works for me on both a real device and a emulator. Not sure what is the problem you are having, the cert file I have is the same format. Let's try it again together later and see if we can figure it out.

PR #154, opened 2018-07-02
toc poc for security service only

finp:
@johnfriz  This shows a minimal toc, could be used for any/all nodes
 

screenshot from 2018-07-02 21-09-25


johnfriz:
Discussed with @finp and have decided to clean up the toc rather than restricting to 1 level

PR #148, opened 2018-07-02
removing next steps and pre-reqs

laurafitzgerald:
**JIRA**
https://issues.jboss.org/browse/AEROGEAR-3513

**Tasks**
- [x] Remove next steps from getting-started
- [x] Remove pre reqs from getting-started

The removal of the `Next Steps` from IDM section in getting started will be resolved by the change in [another pr](https://github.com/aerogear/mobile-docs/pull/143) but I've still made the change to the file so it's consistent with the others. There's a couple of places where the file is linked to which will also lose the next steps at the end but I believe this will be resolved by the work to only have links to the 12 apostles.

@finp

PR #158, opened 2018-07-03
Cleaning up Identity Management docs

johnfriz:
- Removed (most) pre-reqs (except where they made sense) and all Next Steps
- Reordered section for a better flow
- Improved nesting of sub items within Setup
- Removed include of attributes file in generic-binding.adoc as it was not rendering the `{service-name}` param that was set by the file that was importing the generic-binding.adoc - the attributes file was resetting it to blank
 
TOC Before:

screenshot 2018-07-03 01 04 47

TOC After: 

screenshot 2018-07-03 01 03 54


finp:
looking now

laurafitzgerald:
@johnfriz I also made some changes in [this pr](https://github.com/aerogear/mobile-docs/pull/148) that affects your changes here.

PR #159, opened 2018-07-03
AEROGEAR-3526 - IDM step review

ciaranRoche:
##JIRA
https://issues.jboss.org/browse/AEROGEAR-3526

## Additional Information
Removed an unnecessary step as the `create binding` is visible in the mobile service list, it is only after a service bind is the `>` icon visible.

In the Keycloak config, we tell the developer to add a wildcard to the web origins, this caused problems for us in the cordova showcase app. If we are to follow the config we have in `keycloak.security.feedhenry.org` setup, we have the schema (`org.aerogear.js.showcase`) added to the web origin and in the redirect uri's we have schema followed by `:/callback` added (`org.aerogear.js.showcase:/callback`).

Shall we follow the conventions we have in `keycloak.security.feedhenry.org`?

### DIFF
--- a/modules/ROOT/pages/keycloak/keycloak-setup.adoc
+++ b/modules/ROOT/pages/keycloak/keycloak-setup.adoc
@@ -40,9 +40,5 @@ include::{partialsdir}/generic-service-nav.adoc[]
 . Open `Clients` from the menu.
 
 . Open your Android client.
-
-. Add the schema selected in the step 'Choose the schema of a redirect url' above to `Valid Redirect URIs`.
-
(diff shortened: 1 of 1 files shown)


wei-lee:
@ciaranRoche Yes, I think so.

ciaranRoche:
@wei-lee I have changed that, is that wording ok?
keycloak-conf-2

wei-lee:
@ciaranRoche looks good. The only suggestion I have is to add a link to Keycloak about how to configure a client so that if a user needs to know more, they will know where to find them.

PR #166, opened 2018-07-03
update the layout in the api ref doc

wei-lee:
Some of the links are still not working, will need to update them once the right links are available. 

But want to get some feedback on the layout.

screen shot 2018-07-03 at 13 29 03

@johnfriz @finp Wdyt?


wei-lee:
screen shot 2018-07-03 at 15 19 38

PR #168, opened 2018-07-03
Fixed an error when publishing

jasonmadigan:
Unsure if it's just me, but when running `./bin/publish.sh`, I get an error when cloning `mobile-docs`:

```
error: callback returned unsupported credentials type: git@github.com:aerogear/mobile-docs.git
```

Using an HTTP link instead works fine. May be a macOS issue? This should use work everything.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -6,7 +6,7 @@ site:
   start_page: aerogear::getting-started
 content:
   sources:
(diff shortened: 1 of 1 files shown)


jasonmadigan:
Spotted this an error when publishing. No big deal, but this seems to fix.

PR #175, opened 2018-07-03
AEROGEAR - 3555 -add mobile client downloading and installing app steps

ciaranRoche:
## JIRA
https://issues.jboss.org/browse/AEROGEAR-3555

## SCREENSHOT
downloading-installing-apps

### DIFF
--- a/modules/ROOT/pages/_partials/mobile-client-downloading-and-installing-apps.adoc
+++ b/modules/ROOT/pages/_partials/mobile-client-downloading-and-installing-apps.adoc
@@ -1,0 +1,10 @@
+[discrete]
+== Downloading and Installing Apps:
+
+Once the build has completed you can then press the `Download` button and you will be presented with a QR code. Applications to scan QR codes are available in Google Play for Android or the App Store for iPhone. Once the QR code has been scanned it will open the download URL in the device and the application will download.
+
+image:mobile-clients-builds-complete.png[Available Mobile Clients]
+
+If you do not have a QR code scanner you can also download the artifact from the URL in the openshift web console. Once downloaded connect a device to your local machine, then open XCode and navigate to Window > Devices and Simulators. Here you can install your mobile application by pressing `+` button at the bottom, then choose your download IPA file for installation. 
+
+NOTE: For Android devices you need to follow the same steps, just using Android Studio to install the APK file on your device or AVD.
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/_partials/mobile-client-downloading-and-installing-apps.adoc
```diff
@@ -0,0 +1,10 @@
+[discrete]
+== Downloading and Installing Apps:
+
+Once the build has completed you can then press the `Download` button and you will be presented with a QR code. Applications to scan QR codes are available in Google Play for Android or the App Store for iPhone. Once the QR code has been scanned it will open the download URL in the device and the application will download.
+
+image:mobile-clients-builds-complete.png[Available Mobile Clients]
+
+If you do not have a QR code scanner you can also download the artifact from the URL in the openshift web console. Once downloaded connect a device to your local machine, then open XCode and navigate to Window > Devices and Simulators. Here you can install your mobile application by pressing `+` button at the bottom, then choose your download IPA file for installation.
```
laurafitzgerald: could we use the tab format here for these instructions?

finp: maybe  just make two bullets, one for ios and android
and we save the tabs for the four platforms (rather than the two device types)?
Maybe not a rule, but here where it's just two lines, we're not saving any space on screen

ciaranRoche:
@finp @laurafitzgerald I updated the steps into two bullet points, I have attached a screenshot of how it looks, let me know if you want me to change it to the tab layout

apk-steps

modules/ROOT/pages/_partials/mobile-client-downloading-and-installing-apps.adoc
laurafitzgerald: yes makes sense @finp

laurafitzgerald:
@ciaranRoche can we get this partial added to the Mobile Services Section https://docs.aerogear.org/aerogear/latest/mobile-cicd.html#proc_deploying-a-mobile-app_proc_deploying-a-mobile-app

ciaranRoche:
@laurafitzgerald made that change, was that the only bit you wanted to be added?

laurafitzgerald:
@ciaranRoche There's a section in Mobile CICD where we are talking about downloading built apps. It could be pulling from another section but this new change is better so could we get that to pull in this new content instead? I left a link in my previous comment.

laurafitzgerald:
Actually on second thought let's get this change merged as there's pre-reqs to be dealt with in the mobile cicd section. @johnfriz we need to create some tasks about removing the pre reqs/next steps from the Service pages now that they are merged into single sections in the same way that getting started was. @ciaranRoche feel free to merge this one as is.

PR #176, opened 2018-07-03
Aerogear 3557 mobile client builds

wei-lee:
ping @aidenkeating @johnfriz @finp


wei-lee:
@johnfriz I think that's a good idea. Those 2 section are should stay together I think. Will update this PR.

wei-lee:
screen shot 2018-07-03 at 22 36 57


@johnfriz @finp updated

PR #204, opened 2018-07-09
feat: initial provisioning and binding docs for sync

darahayes:
This PR introduces the initial docs for Provisioning and Binding the Sync Service.

## Related Tickets:

* https://issues.jboss.org/browse/AEROGEAR-3432
* https://issues.jboss.org/browse/AEROGEAR-3429

## Updated Table in nav-matrix.adoc

screen shot 2018-07-09 at 14 53 24

As you can see there is a row for the sync service. There are working links for everything except the sync-ui. This is to be addressed as part of another ticket.

## Initial Sync Provisioning Doc with correct parameters

I had to zoom out in order to screenshot the full page

screen shot 2018-07-09 at 14 57 44

## Skeleton Intro Doc

screen shot 2018-07-09 at 14 53 34

I didn't provide a useful description of the Sync Service yet because I don't feel like I can describe it well enough right now. (Still under a lot of work). I invite someone to help me describe it!

## Initial Sync Binding Doc

screen shot 2018-07-09 at 14 57 57

@finp for some reason the `{service-name}` variable is not working only in the 'Next Steps' section. The text `Adding Service to your Application` should be `Adding Data Sync Service to your Application.` The variable is defined and is being rendered correctly in other parts of the same page, such as the tile. I have no idea why this is happening.


pb82:
👀

modules/ROOT/pages/sync/provisioning.adoc
```diff
@@ -9,3 +9,29 @@ include::{partialsdir}/attributes.adoc[]
...
+If the Postgres password was generated, retrieve the password using:
+[source,bash]
+----
+oc describe dc postgres-data-sync -n <myproject> | grep POSTGRESQL_PASSWORD
+----
+====
+
+include::{partialsdir}/generic-provisioning-end.adoc[]
+
+== Next Steps
```
pb82: @finp do we really need next steps here? Because they refer to the binding steps which are directly following in the document.

modules/ROOT/pages/sync/binding.adoc
```diff
@@ -5,3 +5,7 @@ include::{partialsdir}/attributes.adoc[]
 = Binding a {mobile-client} with the {service-name} Service
 
 include::{partialsdir}/generic-binding.adoc[]
+
+== Next Steps
+
+* xref:sync/adding-sync-to-your-app.adoc[Adding {service-name} Service to your Application]
```
pb82: `service-name` is undefined here, you need to either add `:service-name: Data Sync` to the top of the document or use `:sync-service:` which is defined in the main attributes.adoc.

pb82:
Look good in general, i added a few comments. Leaving it up to @finp to merge.

modules/ROOT/pages/sync/binding.adoc
darahayes: @pb82 I don't understand. I have `:service-name: Data Sync` at the top of the document in the exact same way as the other docs.

modules/ROOT/pages/sync/provisioning.adoc
darahayes: @pb82 This is how it's done in the other documents.

modules/ROOT/pages/sync/binding.adoc
pb82: @darahayes Strange, when i render it it shows up as this:

bildschirmfoto 2018-07-10 um 12 02 54

But service-name is defined

pb82: It looks like using an include after defining a variable makes the variable undefined after the include. To fix this you can use `Adding {sync-service} Service` instead.

aliok:
Looks good to me. There is a general problem in the left navigation I think. Not introduced by your PR.
Maybe it is because we need to rebase onto master. I am afraid we're gonna end up touching lots of things if we wait too long.

darahayes:
@pb82 @finp - The issue with the 'Sync Service' variable not showing up is now fixed. Are we good to merge?

PR #208, opened 2018-07-11
Correct scheme setup information

wtrocki:
Linked change: https://github.com/aerogear/ios-showcase-template/pull/52


wtrocki:
@ciaranRoche It's confusing as :/callback is part of the scheme.
Explained that with details in linked PR.

PR #210, opened 2018-07-11
Remove duplicate certs docs

aidenkeating:
@finp The cert information was already included in a partial. This removes the additional one and ensures it's still referenced properly from the nav


aidenkeating:
@finp Updated

aidenkeating:
@finp Rebased

PR #215, opened 2018-07-11
move section to end

laurafitzgerald:
**What**
Move ShowCase App Features section to the bottom of the page
Move navigation item to match

@finp


laurafitzgerald:
@finp @aidenkeating header updated

PR #227, opened 2018-07-12
Bump required Ansible version to 2.6

matskiv:
Installer requires Ansible 2.6, see:
https://github.com/aerogear/mobile-core/blob/master/installer/install.sh#L118-L121


jasonmadigan:
@matskiv 👍

PR #279, opened 2018-07-13
AEROGEAR-7541 - Add fedora firewall options

ciaranRoche:
## JIRA
JIRA - https://issues.jboss.org/browse/AEROGEAR-7541

## Description

While investigating the above bug it was found that on Fedora all ports under 1025 are blocked, this meant that anyone running a local cluster on Fedora would not be able to interact with any services that used a port under 1025.

As it stands the documentation gets a Linux user to expose a port on their `dockerc` this on its own is not sufficient. 

This change adds a warning to the pre-req informing a Linux user of the potential issue and gives a solution. 

## Screenshot
screen

## Note
I have created an issue based on this PR to potentially add a troubleshooting section https://github.com/aerogear/mobile-docs/issues/280

### DIFF
--- a/modules/ROOT/pages/_partials/installing-mobile-services.adoc
+++ b/modules/ROOT/pages/_partials/installing-mobile-services.adoc
@@ -29,14 +29,6 @@ $ firewall-cmd --permanent --zone dockerc --add-port 443/tcp
 $ firewall-cmd --reload
 ----
 
-WARNING: Due to a number of services using ports which on Linux are restricted to root users only. This may result in issues if you try to interact with services on a local cluster from an external device. If you encounter these issues, you can add the ports to your firewall. Depending on the port your service uses, an example of the ports you may want to add to your firewall are:
-
-----
-$ firewall-cmd --add-port 443/tcp
-$ firewall-cmd --add-port 80/tcp
-----
-
-
 [[procedure]]
 == Procedure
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/installing-mobile-services.adoc
```diff
@@ -29,6 +29,11 @@ $ firewall-cmd --permanent --zone dockerc --add-port 443/tcp
 $ firewall-cmd --reload
 ----
 
+WARNING: Firewall issues may arise on Linux if you try to interact with services on a local cluster from an external device. In order to overcome these issues, you can add the following ports to your firewall.
```
StephenCoady: I would add something along the lines of "depending on the port your service uses, an example of the ports you may want to add to your firewall are:"

ciaranRoche:
@StephenCoady made that change if you could have another look, thanks
firewall

modules/ROOT/pages/_partials/installing-mobile-services.adoc
```diff
@@ -29,6 +29,12 @@ $ firewall-cmd --permanent --zone dockerc --add-port 443/tcp
 $ firewall-cmd --reload
 ----
 
+WARNING: Firewall issues may arise on Linux if you try to interact with services on a local cluster from an external device. In order to overcome these issues, you can add the ports to your firewall. Depending on the port your service uses, an example of the ports you may want to add to your firewall are
```
finp: isn't this already in the text above?

ciaranRoche: That is for adding a port to the DockerRc, this change is to flag a seperate issue with the Linux firewall, and it is only a problem if you are using a localcluster on Linux with an app on an external device. 
Because of that I raised this issue about a troubleshooting section https://github.com/aerogear/mobile-docs/issues/280, as pieces like this would be more suited to a troubleshooting guide then in the docs as a warning.

ciaranRoche:
@finp I made changes to the paragraph adding some context about restricted ports, if you wouldn't mind having another look!

modules/ROOT/pages/_partials/installing-mobile-services.adoc
```diff
@@ -29,6 +29,12 @@ $ firewall-cmd --permanent --zone dockerc --add-port 443/tcp
 $ firewall-cmd --reload
 ----
 
+WARNING: Due to a number of services using ports which on Linux are restricted to root users only. This may result in issues if you try to interact with services on a local cluster from an external device. If you encounter these issues, you can add the ports to your firewall. Depending on the port your service uses, an example of the ports you may want to add to your firewall are:
```
finp: Some {product-name} use ports that are restricted on Linux. If you encounter issues connecting to {product-name} from external devices, add the required ports to your firewall. For example:

ciaranRoche:
@finp 
trouble

PR #281, opened 2018-07-13
added docs for datasource view

cfoskin:
Ping @pb82 @josemigallas  

Ive added some docs for the datasource view on the UI.
Please let me know what you think


modules/ROOT/pages/sync/data_sources.adoc
```diff
@@ -0,0 +1,58 @@
+include::{partialsdir}/attributes.adoc[]
+
+//':context:' is a vital parameter. See: http://asciidoctor.org/docs/user-manual/#include-multiple
+:context: data_sources
+
+[id='{context}_data_sources']
+= Managing your data sources using the {sync-service} UI
```
pb82: This does currently not appear in nav.adoc, should it? Or maybe add the content directly to data_sync_ui.adoc?

modules/ROOT/pages/sync/data_sources.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+== Data Source Introduction
+
+A Data Source is a representation of your data storage solution. The Data Source consists of a *Name*, a *Type*, and a *Configuration*. The *Type* of Data Source which are currently supported are *Postgres* and *InMemory*. For information on how to use *Postgres*, refer to the documentation https://www.postgresql.org/docs[here]. The *InMemory* Data Source that is currently supported is *NeDB*, for more information on how to use this please refer to the the documentation https://github.com/louischatriot/nedb[here]. *Note* that the *InMemory* Data Source is not recommended for production use.
```
pb82: InMemory -> In Memory

pb82: Postgres is not currently supported but i think it's ok to mention it here already as it will be the next data source to add.

modules/ROOT/pages/sync/data_sources.adoc
cfoskin: Which would you recommend?

modules/ROOT/pages/sync/data_sources.adoc
cfoskin: yep, I added it based on the fact we will be doing it

modules/ROOT/pages/sync/data_sources.adoc
pb82: I think we should have one one document for the whole section and use anchors to link to specific sections from nav.adoc.

modules/ROOT/pages/sync/data_sources.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+image::datasourceOptions.png[Choose Edit Opion]
+
+Update the data source with the new values.
+
+image::updateDatasource.png[Update Data Source]
+
+=== Delete an Existing Data Source
+
+To delete a data source, choose the *Delete Data Source* option from the data source menu. Next confirm that you wish to delete the data source.
```
pb82: Let's add here that when you delete a data source all resolvers that are attached to it will also be deleted.

pb82: Maybe as a `NOTE:`

modules/ROOT/pages/sync/data_sources.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+image::deleteDatasource.png[Delete Data Source]
+
+=== Create a New Data Source
+
+If no data source exists, a new data source can be created by clicking the *Add Data Source* button
+
+image::createDatasource.png[Create New Data Source]
+
+Enter the *name* of the data source into the required fields on the modal. Next choose the *type* of the data source from the options provided. Next choose if you would like to *timestamp* the data source.
```
pb82: `Next choose if you would like to *timestamp* the data source.` -> `Next choose if you would like to add timestamps to your records.`

modules/ROOT/pages/sync/data_sources.adoc
```diff
@@ -0,0 +1,58 @@
...
+
+To delete a data source, choose the *Delete Data Source* option from the data source menu. Next confirm that you wish to delete the data source.
+
+image::deleteDatasource.png[Delete Data Source]
+
+=== Create a New Data Source
+
+If no data source exists, a new data source can be created by clicking the *Add Data Source* button
+
+image::createDatasource.png[Create New Data Source]
```
finp: does this have any value?

PR #283, opened 2018-07-16
Add step to Android Self Signed Certs

ciaranRoche:
## Description
Currently, we do not have documented that we can push a certificate to a device using Android Debug Bridge. This method would be the quickest way for a developer to get a certificate onto a device.

## Screenshot
command

### DIFF
--- a/modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
+++ b/modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
@@ -51,6 +51,11 @@ Drive which is accessible from the device, browse to the certificate and
 proceed to download it.
 * USB - Attach the device to the machine hosting
 OpenShift via USB and drag the certificate to a devices file system. Here is an https://support.google.com/nexus/answer/2840804?hl=en[example guide for Google Nexus devices]. It may be different for other devices.
+* Android Debug Bridge (adb) - Use the adb push command to push the certificate to device or emulator:
++
+----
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/proc_self-signed-certs.adoc
```diff
@@ -51,6 +51,11 @@ Drive which is accessible from the device, browse to the certificate and
 proceed to download it.
 * USB - Attach the device to the machine hosting
 OpenShift via USB and drag the certificate to a devices file system. Here is an https://support.google.com/nexus/answer/2840804?hl=en[example guide for Google Nexus devices]. It may be different for other devices.
+* Android Debug Bridge (adb) - Use the adb push command to push the certificate to device or emulator:
```
finp: this looks like a good change, but does it open a can of worms.. how do i use adb, why didn't you mention it earlier, do I need to use adb? etc

What would work is if we had a better link above to a proper google article, that would make this 'adb' bit seem like a natural next step

ciaranRoche: Good point, as we have already called out in the documentation about setting up your environment, a developer will get adb out of the box when installing android studio.
Would making Android Debug Bridge a clickable link to some google documentation be sufficient?

finp: @ciaranRoche  yeah, go for it

PR #293, opened 2018-07-18
remove next steps and add terminology

pb82:
Removed the next steps. The sync docs now behave like one big document that you can srcoll from top to bottom and has anchors in nav.adoc.

I've also made a stab at the sync service intro and added some terminology.

### DIFF
--- a/modules/ROOT/pages/_partials/generic-service-intros.adoc
+++ b/modules/ROOT/pages/_partials/generic-service-intros.adoc
@@ -93,6 +93,11 @@ NOTE: {device-security-service} is not currently associated with an OpenShift se
 = {sync-service}
 //tag::sync-service[]
 
-TODO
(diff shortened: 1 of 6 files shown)


laurafitzgerald:
@pb82 thanks for the pr. is the plan to keep the sync docs in it's own branch until the service is completed?
cc @finp I've seen cases with long standing branches where the merge/rebase becomes a bit tricky so could be valuable to rebase on master now before it progresses further?

finp:
I'd hope that rebasing would be easy, that we have structured the docs to facilitate that use case. But +1 Laura, rebase early, rebase often

pb82:
@laurafitzgerald Yes, i wouldn't merge this to master until we have a first version released. Agree that rebasing from time to time makes a lot of sense.

Issue #282, opened 2018-07-16
heading levels in Mobile CI/CD


pmacko1:
https://docs.aerogear.org/aerogear/latest/mobile-cicd.html#provisioning-mobile-ci-cd-service

do we want to have this as level 0 heading? it builds and works but antora shouts that it is an error

pmacko1:
fixed in #295

PR #295, opened 2018-07-18
addressing #282

finp:
fixed mobile ci/cd issue #282 
seems to have been related to a missing newline rather than level


pmacko1:
@finp it did fix the issue. Idk how or why, it just did.

PR #330, opened 2018-08-01
Gdoc idm review - Aiden

finp:
Based on comments from @aidenkeating  in gdoc


modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -120,6 +120,12 @@ IAuthService authService = MobileCore.Instance.GetInstance<IAuthService>();
 var authOptions = DependencyService.Get<IAuthenticateOptionsProvider>().GetOptions();
 authService.Authenticate(authOptions).ContinueWith(user =>
 {
+    // This is a Task
```
aidenkeating: I think we can simplify this to.

```
// Get the authenticated User from the provided Task.
Console.WriteLine(user.Result);
```

Issue #250, opened 2018-07-12
getting started - configuring the service


mmurphy:
steps 6 and 7
it is not clear that <schema> refers to the schema name referenced above, or if it's the text "<schema>"

Additionally it is not clear if the 2 items should be added as additional URLs or if they should be added to the existing URLS

finp:
1. adding example to make this clearer, handling in https://github.com/aerogear/mobile-docs/issues/226
2. We ask users to 
 Add `<schema>:/callback` to `Valid Redirect URIs`.
@ciaranRoche  can you confirm that we're asking users to add these as additional urls?

ciaranRoche:
Yes, for the `Valid Redirect URIs` we need to add `:/callback` to the `schema`. For the `web origins` it is just the `schema`.
We have an example `schema` in the dialog box above the steps. I believe we wanted to keep the schema example separate to this step to avoid people copy and pasting directly.

Issue #226, opened 2018-07-12
in the "Configuring the Service" section of the Doc's


Nanyte25:
We need some extra steps under "Configuring keycloak" https://docs.aerogear.org/aerogear/latest/getting-started.html#prerequisites

Open Clients from the menu.

Open your client. The name of your client is derived from the name of the Mobile Client, the name of the mobile development platform and the client type, for example myapp-android-public.

User needs to click into "myapp-android-public" under "Client ID's"

Add <schema>:/callback to Valid Redirect URIs.
take the example schem and use it in the case of android?

Add com.aerogear.androidshowcase:/callback to Valid Redirect URIs?

Add <schema> to Web Origins.

finp:
. Open your client. The name of your client is derived from the name of the Mobile Client, the name of the mobile development platform and the client type, for example myapp-android-public.

. Select the client, for example, click myapp-android-public.

. Add <schema>:/callback to Valid Redirect URIs. For example, if you are using Android, com.aerogear.androidshowcase:/callback

. Add <schema> to Web Origins. For example, if you are using iOS, org.aerogear.ios-showcase-template

. Save your changes.

PR #334, opened 2018-08-02
issue 226 and 250

finp:
update keycloak process


finp:
addresses #226 and #250

PR #335, opened 2018-08-02
add docs for data sources

pb82:
Docs for the In Mem and Postgres data sources.

### DIFF
--- a/modules/ROOT/pages/sync/data_sources_ui.adoc
+++ b/modules/ROOT/pages/sync/data_sources_ui.adoc
@@ -56,12 +56,14 @@ For more information about the _In Memory_ option have a look at xref:#in-memory
 
 You need to provide a name for the data source. The only other option is `TimeStamp Data`. If checked, timestamps will be stored along with your data in `NeDB`.
 
+WARNING: Some features like unique indices and persistence are not supported in the In Memory database and it is therefore only recommended for evaluation purposes.
+
 [[postgres-option]]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/sync/data_sources_ui.adoc
```diff
@@ -48,6 +48,23 @@ NOTE: When you delete a data source, all resolvers that are attached to it will
...
+=== In Memory Data Sources
+
+You need to provide a name for the data source. The only other option is `TimeStamp Data`. If checked, timestamps will be stored along with your data in `NeDB`.
+
+[[postgres-option]]
+=== Postgres Data Sources
+
+You need to provide a name for the data source. `Postgres` has a number of other required options:
+
+. _Server_: The URL where the PostgreSQL server is located.
```
aliok: Isn't this actually a "hostname"?

aliok:
I just remembered this thread: https://groups.google.com/forum/#!searchin/aerogear/data$20sync$20unique%7Csort:date/aerogear/mb1zGv_5n20/-WGJSiRsAwAJ

We need to document InMem db doesn't support unique indices. I think this is the right place to document that. In addition to "The In Memory Data Source is not recommended for production use." we can say something like "some features like unique indices, persistence, etc. is not supported on InMem data source." What do you think?

aliok:
looks good other than the minor comments above

PR #342, opened 2018-08-03
Sync dashboard explanation

aliok:
Documented the sync dashboard's current state : [1]

Created a single adoc file and included it in 2 places (same as Keycloak).
1. "Mobile Metrics" -> "Monitoring Data Sync Service"
2. "Sync Service" -> "Monitoring Data Sync Service"

[1]: https://groups.google.com/forum/#!topic/aerogear/OC122HNWNfo

PR #343, opened 2018-08-03
Aerogear 7648 docs metrics dashboard customization

aliok:
Depends on https://github.com/aerogear/mobile-docs/pull/342

We already had a document fragment that explains how to use and extend Grafana dashboards, but it was not included anywhere. I included it in the metrics page now.


aliok:
@pb82 now this PR is cleaner. ready for review.

PR #344, opened 2018-08-06
add resolver mapping docs

pb82:
Docs for Schema Editor and Resolver Mappings


modules/ROOT/pages/sync/resolvers_ui.adoc
```diff
@@ -0,0 +1,88 @@
...
+context: {
+    arguments,    // The arguments passed to the resolver
+    parent        // Parent object reference
+}
+```
+
+For example, a request mapping to retrieve all the comments for a particular `Note` could look like this:
+
+```
+select * from Comments where NoteId = '{{ context.parent.id }}'
```
aliok: Is this the place to document mapping formats for different data source types?

You know, use NeDB queries/operations for Inmem, use SQL for Postgres. We can refer to documents of them, but maybe give 1 example each?

josemigallas:
LGTM for what it's worth

PR #362, opened 2018-08-15
Decouple service creation from MobileCore

danielpassos:


modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -138,11 +135,10 @@ The following section describes how to perform a logout.
 .Android
 ****
 
-. Get an instance of the auth service and retrieve the current user:
+. Retrieve the current user:
 +
 [source,java]
 ----
-AuthService authService = MobileCore.getInstance().getService(AuthService.class);
 UserPrincipal currentUser = authService.currentUser();
```
wtrocki: Question. Do we need to provide info how authService if being created here. Old snippet done that.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -29,12 +29,10 @@ Callback authCallback = new Callback<UserPrincipal>() {
 };
 ----
 
-. Retrieve an instance of the `AuthService` and use the `login` method to hand over to the login browser:
+. Use the `login` method to hand over to the login browser:
 +
 [source,java]
 ----
-AuthService authService = MobileCore.getInstance().getService(AuthService.class);
```
wtrocki: After removing this code is not going to `compile` I know it's snippet but best to still create this service from singleton.

modules/ROOT/pages/keycloak/authenticating.adoc
danielpassos: There is a section called `Initializing the SDK` with instruction how to ceate an `authService`

modules/ROOT/pages/keycloak/authenticating.adoc
danielpassos: There is no singleton because `AuthService` creation requires an `AuthServiceConfiguration` provides by the developer.

finp:
@wtrocki @danielpassos  ok to merge?

wtrocki:
We need that to reflect current state of the SDK. 
@danielpassos if this was released in the latest version feel free to merge it.

danielpassos:
@finp The sync branch is going to be merge soon-ish so it does not make sense any more

PR #401, opened 2018-08-30
Added Android and iOS examples for sync and auth

jhellar:
# Motivation

https://issues.jboss.org/browse/AEROGEAR-7812


modules/ROOT/pages/sync/authentication.adoc
```diff
@@ -0,0 +1,25 @@
+include::{partialsdir}/attributes.adoc[]
+
+= Authentication for {sync-service} service
+
+== Using the {sync-service} SDK with {keycloak-service} SDK
```
wtrocki: {sync-service} Authentication ?

modules/ROOT/pages/sync/authentication.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+When user login in with {keycloak-service} SDK, user principal can be used to retrieve token. {sync-service} SDK can use this token to preform requests to the {sync-service} backend.
```
wtrocki: This is really raw explanation. As mentione in tickets we need to provide numer of steps. 
- Integrate with Auth SDK (wityh link)
-  What is needed for sync etc.

I know I said that this will be simple, but it's too simple.

psturc: preform -> perform

modules/ROOT/pages/sync/authentication.adoc
finp: +1 we should describe the benefit to the user not the technologies, especially in headings

PR #402, opened 2018-08-30
added custom scalar warning

StephenCoady:
https://issues.jboss.org/browse/AEROGEAR-7717

### DIFF
--- a/modules/ROOT/pages/sync/schema_ui.adoc
+++ b/modules/ROOT/pages/sync/schema_ui.adoc
@@ -34,7 +34,7 @@ type Note {
 Save the schema py pressing the `Save Schema` button. Upon saving the {sync-service} will check your schema for errors and compile it to a JSON representation.
 You can download this JSON representation by clicking on `Download Compiled Schema`. This is needed in the client SDKs.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/sync/schema_ui.adoc
```diff
@@ -34,6 +34,8 @@ type Note {
 Save the schema py pressing the `Save Schema` button. Upon saving the {sync-service} will check your schema for errors and compile it to a JSON representation.
 You can download this JSON representation by clicking on `Download Compiled Schema`. This is needed in the client SDKs.
 
+NOTE: When defining custom scalars on your schema it is important to assert that the data types are what you are actually expecting. For example if passing a JSON object consisting of String fields, it is important to validate they are actually Strings.
```
wtrocki: `scalars` could have link to graphql docs.
data types are what you are actually expecting. => `matching expected type`

`For example if passing a JSON object consisting of String fields, it is important to validate they are actually Strings.` Example gives little context. 
This is actually resolver specific (and client). I will skip that part completely and say that Scalar definitions could potentially lead to SQLInjection etc.

PR #415, opened 2018-08-30
414 fix method name

pmacko1:
merge if ok pls


finp:
@jhellar  thanks for merging, but in future, leave it to docs to merge

pmacko1:
@finp is ok, I asked Jan to merge if correct.

PR #416, opened 2018-08-30
Sync authorization docs

darahayes:
The purpose of this document:

* Directs the user to the appropriate keycloak documentation for configuring realms, client and realm roles.
* Explains `@hasRole` directive usage in detail.
* Gives a real world example of how the `@hasRole` directive might be used in a schema.

### DIFF
--- a/modules/ROOT/pages/sync/authorization.adoc
+++ b/modules/ROOT/pages/sync/authorization.adoc
@@ -1,0 +1,83 @@
+include::{partialsdir}/attributes.adoc[]
+
+:service-name: Data Sync
+
+= Role Based Authorization in {sync-service}
+
+== Prerequisites
+
+* xref:./binding_to_keycloak.adoc[Binding {sync-service} to Keycloak].
+* link:https://www.keycloak.org/docs/3.4/getting_started/index.html#creating-a-realm-and-user[Creating a Realm in Keycloak].
+* link:https://www.keycloak.org/docs/3.4/authorization_services/index.html#_getting_started_hello_world_enabling_authz_services[Enabling Authorization Services in Keycloak].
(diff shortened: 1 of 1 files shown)


wtrocki:
@darahayes Checked initial version and it makes sense but we may need to make final revisions.

wtrocki:
We may need to join together https://github.com/aerogear/mobile-docs/pull/403 with this PR

modules/ROOT/pages/sync/authorization.adoc
```diff
@@ -0,0 +1,110 @@
...
+* Fields
+* Queries
+* Mutations
+* Subscriptions
+
+The `@hasRole` usage is as follows:
+
+* `@hasRole(role: String)`
+  * Example - `@hasRole(role: "admin])`
+  * If the authenticated user has the role `admin` they will be authorized.
```
wtrocki: [Optional ] the role `admin` => the `admin` role

modules/ROOT/pages/sync/authorization.adoc
```diff
@@ -0,0 +1,110 @@
...
+* `@hasRole(role: [String])`
+  * Example - `@hasRole(role: ["admin", "editor"])`
+  * If the authenticated user has **at least one of the roles** in the list, they will be authorized.
+
+**The default behaviour is to check client roles.** For example, `@hasRole(role: "admin")` will check that a user has a client role called `admin`.
```
wtrocki: This section is no longer actual.

darahayes: @wtrocki it is indeed still true. `hasRole(role: "admin") well check against client roles, whereas `hasRole(role: "realm:admin") will check against a realm role.

wtrocki: ah.. right. Yes. That is true

wtrocki:
Fixed typo. Good to merge!

PR #423, opened 2018-09-04
244 fix keycloak binding procedure

pmacko1:


modules/ROOT/pages/keycloak/binding.adoc
```diff
@@ -5,10 +5,7 @@ include::{partialsdir}/attributes.adoc[]
 = Binding a {mobile-client} 
 
 include::{partialsdir}/generic-binding.adoc[]
-
-. Choose the *Keycloak client type*. The default is `public` which allows the client to request tokens while type `bearer` should be used to bind to services that only need to verify tokens.
-+
-NOTE: The `public` client type should be used to bind to {mobile-client}s because they only need to request tokens, but not verify them. When binding to other services you should use client type `bearer` unless they also need to authenticate.
+NOTE: Use `public` when binding a {mobile-client} to {service-name} Service, as it enables {mobile-client} to request tokens ({mobile-client} doesn't need to verify tokens). Use `bearer` when binding {service-name} Service to other services (If the other service to be bound does not need authentication.).
```
finp: this is only for keycloak, right?
Choose `public` as the *Keycloak client type*  because that enables {mobile-client} to request tokens without verification. 

This piece below needs to move to a new section?

Choose `bearer` when binding {service-name} Service to other services (If the other service to be bound does not need authentication.).

finp: 3. We should leave it cos it's useful info for a user of mobile services

finp: @secondsun  I'm not sure if the text change is correct.
1) we don't want to explain the diff twixt public and bearer, that's in the keycloak docs
2) we do want to recommend a path for users (even if other paths are available)

Maybe:

NOTE: Use `public` when binding a {mobile-client} to a mobile service. When binding mobile services to each other, use `bearer`.

finp: @pmacko1 just got two dev leads to agree to my suggested note above

pmacko1: @finp NOTE: Use `public` when binding a {mobile-client} to a {service-name}. When binding mobile services to each other, use `bearer`.

?

finp: @pmacko1 yes, that's it

secondsun: I like that wording too.

pmacko1:
grand guys, thanks, merging

PR #425, opened 2018-09-04
Restores coding section for metrics service

josemigallas:
JIRA: https://issues.jboss.org/browse/AEROGEAR-3374

Docs are missing a rather critical step for Cordova Metrics SDK, the installation of the metrics plugin.

`coding.adoc` restored to be consistent with the other modules.


aliok:
The content looks good, but the headings are a bit off.

screenshot from 2018-09-05 10-44-29

aliok:
@josemigallas 
> by headings you mean Configuring your... and Download the Mobile, don't you? It looks like that both in Push and Security, on the other it looks different (and actually better) in Identity Management.. 

Yes. that one.

> Maybe we should address that in a different ticket, Or I could do it here as well.

It is up to you :) But, we should fix it.

modules/ROOT/pages/metrics/coding.adoc
```diff
@@ -0,0 +1,16 @@
+include::{partialsdir}/attributes.adoc[]
+
+include::configuring-dev-env-metrics.adoc[]
+
+== Setting up {metrics-service} service SDK
```
finp: we don't use this heading elsewhere

modules/ROOT/pages/nav-matrix.asciidoc
```diff
@@ -40,7 +40,7 @@ After choosing a mobile service, use the following table to navigate to the step
 |xref:metrics/index.adoc[{metrics-service}]
 |xref:metrics/provisioning.adoc[Provisioning {metrics-service}] 
 |xref:metrics/binding.adoc[Binding to {metrics-service}] 
-|xref:metrics/configuring-dev-env-metrics.adoc[Adding Metrics to your App]
+|xref:metrics/coding.adoc[Adding Metrics to your App]
```
finp: fyi: this doc never gets published

modules/ROOT/pages/metrics/coding.adoc
```diff
@@ -0,0 +1,16 @@
...
+
+The {metrics-service} service SDK is included by default when installing any of the components from {SDK}, so Metrics are sent automatically from the mobile app.
```
finp: I think this needs to be more explicit.

I like the 'Configuring your Development Environment for the Push Notifications Service' heading followed by:
* download json
* add library from AeroGear SDK
as  a flow

The alternative approach is in keycloak
* download json
* Add 'auth' to your app

whichever path you go, the instructions should be clear, eg:

. If you already use a component from the {SDK} in your app, you can skip this procedure.
. Install the core component
. If you are developing a Cordova app ....

modules/ROOT/pages/metrics/coding.adoc
finp: dang you and your 'facts'... I've created https://issues.jboss.org/browse/AEROGEAR-7893 address this later ;)

josemigallas:
@finp @pmacko1 I re-wrote the PR and rebased, taking as template security docs. Mind reviewing again?

PR #434, opened 2018-09-05
Small fixes

ziccardi:


ziccardi:
ping @wtrocki

wtrocki:
Late to the party. Thanks for fix!

Issue #442, opened 2018-09-07
[glossary] remove runtime connector and metrics from glossary


pmacko1:
metrics does not have any terminology and runtime connector is deprecated

PR #443, opened 2018-09-07
[glossary] remove metrics and runtime connector from glossary

ghost:


pmacko1:
#442

finp:
@pmacko1 why did we remove metrics?

pmacko1:
@finp there is no glossary in metrics. Nor did metrics/ref_terminology.adoc ever exist-> antora resolved it as "include:: ..."

finp:
@pmacko1  thanks for explanation, got it now

Issue #428, opened 2018-09-04
 [idm, implementing auth log out, xamarin] make the docs generic again(not specific to showcase apps)


pmacko1:
## Description
```Device.BeginInvokeOnMainThread(() => ((RootPage)(App.Current.MainPage)).ChangePage(new GoodbyePage()));```
is specific to a showcase app

## Link to Page/Section where Issue was seen
https://docs.aerogear.org/aerogear/latest/identity-management.html#adding-log-out-option-to-your-mobile-app

## Suggested edit
Replace:
```
Device.BeginInvokeOnMainThread(() => ((RootPage)(App.Current.MainPage)).ChangePage(new GoodbyePage()));
```
with:
```
// user is now logged out
```

PR #466, opened 2018-09-28
428 xamarin idm auth logout - make it not specific to showcase apps

pmacko1:
#428 @jhellar


finp:
comparing this to android,ios,cordova, @jhellar  I think they should all be the same, right?
eg:
// User logged out successfully and local Auth tokens are deleted

jhellar:
@finp yes, better to keep it same for all platforms

pmacko1:
@finp @jhellar howboutnow?

jhellar:
LGTM.

PR #508, opened 2018-11-02
Adding documentation for calling external backend and refreshing tokens

ziccardi:
## What:
- [x] Android documentation
- [x] iOS documentation
- [x] js documentation
- [x] xamarin documentation

### DIFF
--- a/modules/ROOT/pages/keycloak/authenticating.adoc
+++ b/modules/ROOT/pages/keycloak/authenticating.adoc
@@ -270,9 +270,9 @@ var result = await authService.Logout(authService.CurrentUser());
 
 NOTE: To perform *backchannel or federated logouts*, you must enable the Backchannel Logout option for the federated identity provider. More information is available in the Keycloak documentation under link:https://www.keycloak.org/docs/3.3/server_admin/topics/identity-broker/oidc.html[OICD Identity Providers, window="_blank"].
 
-== Using auth with backend
+== Authenticating with a back end service
 
-To authenticate service on existing backend server users need to supply additional token:
+In a typical scenario, you not only want the user to authenticate on the mobile app, you also want that user to authenticate with a back end service. To achieve this, you must add an additional token:
 
 [source]
 ----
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -270,6 +270,35 @@ var result = await authService.Logout(authService.CurrentUser());
...
+
+[role="primary"]
+.Android
+****
+
+. Add the `AuthHeaderProvider` instance provided by the Aerogear SDK to the current list of providers:
++
+[source,java]
+----
+authService.getAuthHeaderProvider()
```
wtrocki: Users will not use core for networking so we just showing the method to get provider. that will always get the fresh token for IOS and Android.

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -270,6 +270,35 @@ var result = await authService.Logout(authService.CurrentUser());
 
 NOTE: To perform *backchannel or federated logouts*, you must enable the Backchannel Logout option for the federated identity provider. More information is available in the Keycloak documentation under link:https://www.keycloak.org/docs/3.3/server_admin/topics/identity-broker/oidc.html[OICD Identity Providers, window="_blank"].
 
+== Using auth with backend
```
finp: == Authenticating with a back end service

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -270,6 +270,35 @@ var result = await authService.Logout(authService.CurrentUser());
...
+
+To authenticate service on existing backend server users need to supply additional token:
```
finp: In a typical scenario, you not only want the user to authenticate on the mobile app, you also want that user to authenticate with a back end service. To achieve this, you must add an additional token:

question: where does the user add this? where do they get the token from?

danielpassos:
@ziccardi I have sent https://github.com/ziccardi/mobile-docs/pull/1 to add the JS info

ziccardi:
@finp I just merged js docs from Passos. Mind giving one final look?

modules/ROOT/pages/keycloak/authenticating.adoc
```diff
@@ -270,6 +270,73 @@ var result = await authService.Logout(authService.CurrentUser());
...
+----
+authService.getAuthHeaderProvider()
+----
+. Plug it into your networking layer.
+****
+
+[role="secondary"]
+.Cordova
+****
+Cordova SDK doesn't provide an HTTP layer so, it also doesn't provide an Auth Header. It's your responsibility to attach that to your HTTP provider.
```
finp: 1. bit ambiguous here, do you mean:
The AeroGear Cordova SDK does not provide ...
or 
The Apache Cordova SDK does not provide ...

2. therefore it cannot provide an Auth Header.

3. You must attach an HTTP layer to the HTTP provider .

Is that demonstrated in the code below? if not, do we need to give them a link?

danielpassos: 1. I mean AeroGear Cordova SDK

3. It depends how HTTP provides you are using but is there is an example after this text

@ziccardi Can you update it with the @finp suggestions?

ziccardi: @danielpassos @finp : rewrote. Mind giving a look?

Issue #492, opened 2018-10-17
remove unnecessary attributes.adoc includes


finp:
To allow for reorg, we need to make sure that attributes.adoc is only included once per html page, currently many partials include this file.
include attributes.adoc only in top level file (12 apostles)

pmacko1:
what about mobile security?

finp:
ignore mob security, assume that for any issues, unless it's specifically mentioned

pmacko1:
:ok:

pmacko1:
if i removed the include from _partials, diff reports no changes in htmls

pmacko1:
if i removed the includes from the subfolders, some of the apostles decided to cause havok.
i.e. parameters/variables/whatever the placeholders are called are unresolved. 

It can be because we render the htmls separately and then link them together?

finp:
@pmacko1  One possible workaround for this would be to try:
1. move all subfolders, eg sync, to _partials
2. change all includes in apostles to include::_partials/<existing-path>
3. test
4. remove include::attributes.adoc
5. test again

pmacko1:
13 apostles actually
* xref:getting-started.adoc[Getting Started]
* xref:mobile-developer-console.adoc[Mobile Developer Console]
* xref:con_services.adoc[Mobile Services]
** xref:identity-management.adoc[{keycloak-service}]

** xref:push-notifications.adoc[{unifiedpush-service}]
** xref:mobile-metrics.adoc[{metrics-service}]
** xref:mobile-cicd.adoc[{mobile-ci-cd-service}]
** xref:data-sync.adoc[{sync-service}]

** xref:device-security.adoc[{device-security-service}]
* xref:showcase-apps.adoc[Showcase Apps]
* xref:ref_api.adoc[API Reference]
* xref:mobile-cli.adoc[Mobile CLI]

* xref:glossary.adoc[Glossary]

pmacko1:
stuck, couldn't render docs properly if i removed `include::attributes.adoc` from docs in _partials

finp:
I think the issues you encountered were due to errors at the start of each master assembly file (apostle).
See #512  I believe this works

PR #512, opened 2018-11-04
Fix attributes

finp:
## Motivation
#492 

## What
attribute include is only required at the start of each master  assembly (apostle).
* make sure it's included at start of each master assembly
* remove includes from all other files

## Why
To support refactoring

## How
moved service/* files into _partials

## Verification Steps
run antora


pmacko1:
ctrl-f in data sync for `Make sure it is sync/provisioning.adoc`

pmacko1:
this doesn't work

finp:
need to try a different approach

finp:
@pmacko1 @jstaffor , thinking about reopening this PR.. When you look at the errors that resulted in this PR, they actually reveal issues that are in the existing published doc.. 
For example, goto https://docs.aerogear.org/aerogear/latest/data-sync.html
and search for 'Make sure it is', the following link brings you to https://docs.aerogear.org/aerogear/latest/sync/provisioning.html which we shouldn't be linking to anyway because it's not a master assembly that is referenced in nav.adoc

pmacko1:
so we should be linking only to masters/xrefs in those masters?

Issue #520, opened 2018-11-12
[identity management] fix xref


pmacko1:
1. `Authentication will be performed on initialization as described keycloak/coding.adoc#initializing-the-sdk. To` in https://docs.aerogear.org/aerogear/latest/identity-management.html#user-auth 1st table, tab Cordova

2. in additional information
`Additional configuration for Android`

pmacko1:
Additional configuration for Android is an orphaned file, @finp.

Issue #519, opened 2018-11-12
[getting started] fix xref


pmacko1:
`Log in to the Administration console using the credentials you specified when Provisioning the service (defaults to admin:admin).`

text renders ok and xref works in https://docs.aerogear.org/aerogear/latest/identity-management.html#configuring-keycloak ...

change to getting.started.adoc#provisioning

Issue #518, opened 2018-11-12
[device security] fix xref


pmacko1:
at https://docs.aerogear.org/aerogear/latest/device-security.html#additional-resources
`Besides the security/index.adoc#ref_terminology_Device%20Security_ref_terminology it`

and add heading "Device security" in glossary.adoc

pmacko1:
~we don't include terminology for device security anywhere~ @finp 
We do, but it misses a heading actually. see https://docs.aerogear.org/aerogear/latest/glossary.html
idk if intentional but i can fix it as well, just shout

PR #526, opened 2018-11-12
520_519_518 fix xrefs and mistakes along the way

pmacko1:
#520 #519 #518

### DIFF
--- a/modules/ROOT/pages/_partials/security/proc_performing-device-checks.adoc
+++ b/modules/ROOT/pages/_partials/security/proc_performing-device-checks.adoc
@@ -476,7 +476,7 @@ Executing multiple checks asynchronously is not supported at the moment for this
 
 === Adding Custom Device Checks
 
-Besides the xref:security/index.adoc#ref_terminology_Device%20Security_ref_terminology[Provided Device Checks] it's also possible to make use of your own custom checks. Follow the next steps depending on your platform to implement them:
+Besides the xref:glossary.adoc#ref_terminology_Device Security_ref_terminology[Provided Device Checks] it's also possible to make use of your own custom checks. Follow the next steps depending on your platform to implement them:
 
 [role="primary"]
 .Android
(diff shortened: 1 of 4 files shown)


aidenkeating:
@finp Was that approval meant to be on this PR? Also, if you want to take that create-your-own-walkthough doc task, that sounds good to me.

modules/ROOT/pages/_partials/security/proc_performing-device-checks.adoc
```diff
@@ -476,7 +476,7 @@ Executing multiple checks asynchronously is not supported at the moment for this
 
 === Adding Custom Device Checks
 
-Besides the xref:security/index.adoc#ref_terminology_Device%20Security_ref_terminology[Provided Device Checks] it's also possible to make use of your own custom checks. Follow the next steps depending on your platform to implement them:
+Besides the xref:glossary.adoc#ref_terminology_Device Security_ref_terminology[Provided Device Checks] it's also possible to make use of your own custom checks. Follow the next steps depending on your platform to implement them:
```
finp: I'm thinking this should either be link: or that there's a glossary section included in security master assembly and the xref would be internal to security master assembly

pmacko1: I see that "Performing Device Trust Checks in your Device" has snippets required to implement the chosen features, while glossary holds info about those features. Idk if this can be written better, what was the motivation behind separating these two?

pmacko1: i think we used to include glossary in the "masters"

pmacko1: is this ok then?

finp: should be link: instead of xref, I think
reason: xref works in antora, but will not work with asciidoctor
but feel free to test

pmacko1: aha ok will do. I thought that there is a problem with the text. will test, if ok ill merge

pmacko1: built glossary and device-security with asciidoctor, it seems to work

Issue #524, opened 2018-11-12
[device security] include glossary in device-security.adoc


pmacko1:
pls approve @finp

finp:
1. Should there be a glossary section included at the top of device security master assembly file?
2. There seems to be a heading missing from https://docs.aerogear.org/aerogear/latest/glossary.html for security

pmacko1:
should there be a glossary in any master file?
2. fixed in #526

pmacko1:
also this is only a `blind` fix of a broken xref.

pmacko1:
oh, this is also fixed in #526

finp:
@pmacko1  I think we'll continue to have a glossary section for each master.
so there's an issue that there isn't one in device security master.

Can we close this issue and/or create a new issue to tackle that

pmacko1:
@finp there isn't a glossary in any of the masters

finp:
I mean terminology section, like https://docs.aerogear.org/aerogear/latest/identity-management.html#ref_terminology_Identity%20Management_ref_terminology

pmacko1:
i would merge this, because it keeps the original text and raise an issue if there should be glossaries included in the masters wdyt

finp:
@pmacko1  yes, i agree
issue is that we have terminology section in each service master, except device security

pmacko1:
yeah, i see that now. motivation to duplicate the glossaries?

also the device security is different right, it is not a "copy"

pmacko1:
or, this issue is called add xref to glossary, why open a new one, right

PR #530, opened 2018-11-16
include glossary to device-security

pmacko1:
#524

### DIFF
--- a/modules/ROOT/pages/_partials/security/index.adoc
+++ b/modules/ROOT/pages/_partials/security/index.adoc
@@ -8,7 +8,7 @@
 = Introduction to the {device-security-service} Service
 
 include::../generic-service-intros.adoc[tags=device-security-service]
+include::./ref_terminology.adoc[leveloffset=+1]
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/security/index.adoc
```diff
@@ -8,7 +8,7 @@
 = Introduction to the {device-security-service} Service
 
 include::../generic-service-intros.adoc[tags=device-security-service]
+include::./ref_terminology.adoc[leveloffset=+1]
 
-[id='{context}_additional-resources-{context}']
```
pmacko1: what are these tags used for?

pmacko1: @finp

finp: from mod docs templates https://redhat-documentation.github.io/modular-docs/

pmacko1: let's add it when we get to that step later?

PR #559, opened 2019-01-14
review comments

finp:
might be free after 3 to discuss


modules/ROOT/pages/_partials/keycloak/13_authenticating-with-a-back-end-service.adoc
```diff
@@ -109,6 +100,11 @@ var result = await authService.Logout(authService.CurrentUser());
 ----
 ****
 
+// tag::excludeDownstream[]
```
finp: @finp remove tag

PR #564, opened 2019-01-16
Security only tab work done for 'tabs' syntax

jstaffor:
Security only tab work done for 'tabs' syntax. Building locally


jstaffor:
@finp - it's easy for me to close this and the others, and create them against "jstaffor/AK-47_ ... " - you good with this?

PR #588, opened 2019-01-28
Copy/paste keycloak_doc

secondsun:
@darahayes Here's a copy of the keycloak docs

### DIFF
--- a/modules/ROOT/pages/_partials/sync/adding_keycloak.adoc
+++ b/modules/ROOT/pages/_partials/sync/adding_keycloak.adoc
@@ -1,0 +1,60 @@
+= Prerequisites
+
+- There is a Keycloak service available.
+- You have provisioned a sync application using our playbook.
+- oc tools must be installed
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/sync/adding_keycloak.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+= Connecting your application to Keycloak
+
+Any application that connects to keycloak must consume a keycloak.json file.  This guide demonstrates how to add a keycloak.json file to your sync application deployment.  It will create the environment variable `KEYCLOAK_CONFIG` that contain the path to keycloak.json.  It is still your application's responsibility to consume this file.  We have provided an link:https://github.com/aerogear/voyager-server/tree/master/examples/keycloak[example project].
+
+= Create a Keycloak Secret
+
+The following command will create a secret called `sync-keycloak-doc`.  It will contain one key, `keycloak` and a value, the text of your keycloak.json file.
+
+```bash
```
darahayes: Hey @secondsun this is actually not valid asciidoc syntax.

It should be something like this:

```asciidoc
[source, bash]
----
echo "hello world"
----
```

PR #589, opened 2019-01-29
fix getting started-keycloak

pmacko1:
@finp, this updates the getting-started with steps for keycloak-on-mdc as well (hopefully)


finp:
@pmacko1  sorry, this looks out of date now

PR #590, opened 2019-01-29
fix codeblock-Cloning the Showcase App

pmacko1:


pmacko1:
@finp it seems to be working

PR #591, opened 2019-01-29
Update cloning-showcase-app.adoc

pmacko1:
@finp it seems to be working, i have removed the previous change by mistake. (do not blindly force push :) )


pmacko1:
valid, @finp?

PR #594, opened 2019-02-01
update link syntax to ^

finp:
## Motivation
link syntax is not portable

## What
change , window="_blank"] to ^]


finp:
@jstaffor  thanks, will merge

PR #596, opened 2019-02-01
Fix idm xrefs

finp:
## Motivation
fix unresolved xrefs for downstream


modules/ROOT/pages/_partials/keycloak/authenticating-with-a-back-end-service.adoc
```diff
@@ -103,6 +103,10 @@ var result = await authService.Logout(authService.CurrentUser());
 
 .Additional resources
 
+// tag::excludeDownstream[]
+
 * xref:android-additional-configuration[Additional configuration for Android]
```
pmacko1: this xref does not seem to work in local render @finp

PR #603, opened 2019-02-06
docs: Cordova code samples reflecting aerogear-js-sdk changes in API

wojta:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8462

## What
Changed code samples.

## Why
Code samples reflect changes from 2.1.0-rc3 aerogear-js-sdk

## How
Changing code samples in Cordova

## Verification Steps
Check all Cordova code samples.

## Checklist:
- [x] Code has been tested locally by PR requester
- [ ] Changes have been successfully verified by another team member 

## Progress

- [x] Initialization code sample
- [x] Auth service
- [x] Metrics service
- [x] Push service
- [x] Security service

### DIFF
--- a/modules/ROOT/pages/_partials/conf-dev-env-cordova.adoc
+++ b/modules/ROOT/pages/_partials/conf-dev-env-cordova.adoc
@@ -15,5 +15,5 @@ const aerogearConfig = {
     // Replace with your own configuration
 };
 
-init(aerogearConfig);
+const app = init(aerogearConfig);
 ----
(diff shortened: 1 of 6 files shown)


modules/ROOT/pages/_partials/conf-dev-env-cordova.adoc
```diff
@@ -15,5 +15,5 @@ const aerogearConfig = {
     // Replace with your own configuration
 };
 
-init(aerogearConfig);
```
wtrocki: IMHO this is just showing that our flow and documentation snippets are really confusing. It's almost impossible to find that info (and I know about it).  Documentation for this part should be included in each individual SDK so I do not need to jump and find how to configure things. Adding as placeholder for myself to improve stuff afetr.

finp: @wtrocki  yes, you're right ;)
we should revisit this change

austincunningham: Agree this needs to be called out in each as our documentation flow doesn't ever get to this point see https://github.com/aerogear/mobile-docs/issues/654

PR #607, opened 2019-02-06
dont merge: nuke mdc docs to see whether netlify preview works

pmacko1:
ssia


pmacko1:
let's do another pr, with only one html affected

PR #610, opened 2019-02-07
AGMS-94 for idm

finp:
fix cordova references and restore 'auth with back end service'

### DIFF
--- a/modules/ROOT/pages/_partials/keycloak/adding-access-control.adoc
+++ b/modules/ROOT/pages/_partials/keycloak/adding-access-control.adoc
@@ -62,10 +62,10 @@ let hasAdminRole = currentUser.hasRealmRole("admin")
 let hasModeratorRole = currentUser.hasClientRole(client: "my_client", role: "moderator")
 ----
 --
-// end::excludeDownstream[]
 Cordova::
 +
 --
+// end::excludeDownstream[]
 Get the realm roles from the auth service:
 
 [source,javascript]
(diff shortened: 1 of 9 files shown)


jstaffor:
The following comments are based on the differences between (a) the content  present in this PR and (b) the work that was present for commit (https://github.com/aerogear/mobile-docs/commit/0f5de26a794eacd8ca151136541f18dcea9a49de)

jstaffor:
1) (b) does not contain the heading "Configuring your development environment for the Identity Management Service"

jstaffor:
(a) "Move mobile-services.json to the following location in your application project:" where (b) has "Follow the platform-specific instructions:"

jstaffor:
3. (a) "To enable users of the mobile app log out from the service." where (b) has "To perform a logout:"

jstaffor:
4. For Cordova tab under text "Add code to authenticate with a back end service:"

**(a) has :**
The AeroGear Cordova SDK doesn’t provide an HTTP layer, therefore it cannot provide an Auth Header Provider. You must manually attach the Bearer header to your HTTP requests.

req.setRequestHeader('Accept', 'application/json');
req.setRequestHeader('Authorization', 'Bearer ' + authService.extract().token);

**(b) has** 
To logout invoke the logout function:

authService.logout().then(() => {
    // User Logged Out Successfully and local Auth tokens were Deleted
});

finp:
it was actually the file https://raw.githubusercontent.com/aerogear/mobile-docs/86f81110c73e3dfa8cd9c1e87f404459c6761b8f/modules/ROOT/pages/_partials/keycloak/authenticating.adoc that needed comparison. The issue was that the 'auth with back end' text described how to do log out. I've verified that the text in this PR is correct, and will merge

PR #617, opened 2019-02-11
Aerogear 8400 configure showcase app

wei-lee:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8400

## Additional Notes

The main purpose of this PR is to get the content of how to provision the showcase sync server into this repo. 

Originally, this section is supposed to be in `Mobile Services` -> `OpenShift APB`. However, I don't think it makes a lot of sense to put the content in here, due to a few reasons:

1. It shouldn't be called `OpenShift APB`. `APB` is too technical, and it is the means rather than the goal. The goal here is to get data sync running on OpenShift. So I changed it to `Running Data Sync on OpenShift`.
2. The section should be generic and relatively high level, and only reference to the sample application if there is a good reason. At the moment, it references to the basic example, but it doesn't explain why. I don't think we need to reference the basic example here, so I removed it completely. I also don't feel like there is a good reason to have the content of setting up the server of the showcase app in this section as well, as it doesn't make sense to just provision the server without running the client app. So I moved it to its own section under `Mobile Services` -> `Data Sync` -> `Running the Sample Application`, with the idea that we will cover both server and client in this one.

However, given that we already have a showcase app section, and that we are merging the old showcase app and the new sync server app, I think the best place is to put this content to be part of the setup of the (new) showcase app. So for now I am just focusing on adding the content to the repo, we will figure out the best place to put it later on.

@finp


modules/ROOT/nav.adoc
```diff
@@ -79,6 +77,7 @@ include::{partialsdir}/attributes.adoc[]
 **** xref:data-sync.adoc#sync-js-client-auth[Authentication and Authorization]
 **** xref:data-sync.adoc#sync-js-client-metrics[Metrics]
 **** xref:data-sync.adoc#sync-js-client-audit-logs[Audit Logs]
+*** xref:data-sync.adoc#sync-sample-app[Running the Sample Application]
```
wei-lee: we probably want to move it to be part of the showcase app later on.

modules/ROOT/nav.adoc
```diff
@@ -64,12 +64,10 @@ include::{partialsdir}/attributes.adoc[]
 **** xref:data-sync.adoc#sync-server-auth[Authentication and Authorization]
 **** xref:data-sync.adoc#sync-server-metrics[Metrics]
 **** xref:data-sync.adoc#sync-server-audit-logs[Audit Logs]
-*** xref:data-sync.adoc#getting-started[OpenShift APB]
+*** xref:data-sync.adoc#getting-started[Running {sync-service} on OpenShift]
```
wei-lee: See my comments for the change of the header

modules/ROOT/pages/_partials/sync/getting-started-showcase-app.adoc
```diff
@@ -1,3 +1,54 @@
-= Running Sample Ionic Showcase App
+= Overview
```
wei-lee: this is the main part that needs to be reviewed

modules/ROOT/pages/data-sync.adoc
```diff
@@ -33,25 +33,17 @@ include::_partials/sync/server-metrics.adoc[leveloffset=2]
...
+== Running {sync-service} on OpenShift
 
 [#sync-server-getting-started-overview]
 include::_partials/sync/getting-started-overview.adoc[leveloffset=2]
 
 [#building-sync-container]
 include::_partials/sync/provisioning.adoc[leveloffset=2]
 
-[#sync-server-example-basic]
-include::_partials/sync/sync-server-example-basic.adoc[leveloffset=2]
```
wei-lee: I don't see the value of having this content here at all.

PR #622, opened 2019-02-13
Real time updates

StephenCoady:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8428

## What
Add real time update instructions to both the server and js-client sections

### DIFF
--- a/modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
+++ b/modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
@@ -21,21 +21,17 @@ paramater in the {sync-client} config object. A DataSyncConfig interface is also
 
 == Setting up a Client to use Subscriptions
 
-To set up a client to use subscriptions simply provide a wsUrl string in the config object. There is also
-a `DataSyncConfig` interface available to use. {sync-client} will automatically
-apply the necessary logic to handle websocket requests.
+To set up a client to use subscriptions simply provide a wsUrl string in the config object.
 
 ```
-const config: DataSyncConfig {
-    ...
+const config = {
     wsUrl: "ws://your_websocket_url"
-    ...
 }
 ```
 Where wsUrl is the full URL of the websocket endpoint of your GraphQL server. Once this object is created, use it to initialise {sync-client} with:
 
 ```
-import { createClient } from '@aerogear/voyager-client';
+const { createClient } = require("@aerogear/voyager-client");
 
 const client = createClient(config)
 ```
@@ -119,7 +115,7 @@ An example of using the provided Cordova Network Status, which is an implementat
 This code should be used to re-run a query upon the client coming back to an online state.
 [source, javascript]
 ----
-    import { CordovaNetworkStatus, NetworkInfo } from '@aerogear/voyager-client';
+    const { CordovaNetworkStatus, NetworkInfo } = require("@aerogear/voyager-client");
     const networkStatus = new CordovaNetworkStatus();
     const tasks = []
 
--- a/modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
+++ b/modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
@@ -1,3 +1,139 @@
 = Realtime Updates
 
-TODO
+Subscriptions are a core concept of GraphQL and one which allow us to enable real time updates of data
+from the server to the client.
+To do this GraphQL Subscriptions utilise websockets to enable clients to subscribe to published changes.
+
+The architecture of websockets is as follows:
+
+* Client connects to websocket server
+* Upon certain events, the server can publish the results of these events to the websocket
+* Any _currently connected_ client to that websocket will receive these results
+* The client can close the connection at any time and will no longer recieve updates
+
+Websockets are a perfect solution for delivering messages to currently active clients. To receive updates the
+client must be currently connected to the websocket server, updates made over this websocket while the client is offline
+will not be consumed by the client. For this use case xref:push-notifications.adoc[Push Notifications] are recommended.
+
+{sync-client} comes with subscription support out of the box including auto-reconnection upon device restart or network reconnect.
+To enable subscriptions on your client set the following
+paramater in the {sync-client} config object. A DataSyncConfig interface is also available from {sync-client} if you wish to use it.
+
+== Setting up a Client to use Subscriptions
+
+To set up a client to use subscriptions simply provide a wsUrl string in the config object. There is also
+a `DataSyncConfig` interface available to use. {sync-client} will automatically
+apply the necessary logic to handle websocket requests.
+
+```
+const config: DataSyncConfig {
+    ...
+    wsUrl: "ws://your_websocket_url"
+    ...
+}
+```
+Where wsUrl is the full URL of the websocket endpoint of your GraphQL server. Once this object is created, use it to initialise {sync-client} with:
+
+```
+import { createClient } from '@aerogear/voyager-client';
+
+const client = createClient(config)
+```
+
+== Using Subscriptions
+
+Once the client is configured we can then make use of subscriptions. There are many ways to do this, however some are
+preferred over others.
+
+A standard flow to utilise subscriptions is as follows:
+
+1. Make a network query to get data from the server
+1. Watch the cache for changes to queries
+1. Subscribe to changes pushed from the server
+1. Unsubscibe when leaving the view where there is an active subscription
+
+We can see this happening in the three examples below. `subscribeToMore` ensures that any further updates received from the server force the updateQuery function to be called with `subscriptionData` from the server. Using `subscribeToMore` we can ensure the cache is easily updated as all GraphQL
+queries are automatically notified. For more information please see: link:https://www.apollographql.com/docs/angular/features/subscriptions.html#subscribe-to-more[subscribeToMore Docs].
+
+[source,javascript]
+----
+getTasks() {
+  const tasks = client.watchQuery({
+    query: GET_TASKS
+  });
+
+  tasks.subscribeToMore({
+    document: TASK_ADDED_SUBSCRIPTION,
+    updateQuery: (prev, { subscriptionData }) => {
+      if(subscribtionData.data){
+        const newTask = subscribtionData.data.taskAdded;
+        if (allTasks.find(task) => task.id === createTask.id)) {
+          return prev;
+        } else {
+          return allTasks: [
+              ...prev,
+              newTask
+            ];
+          }
+        } else {
+          return prev
+        }
+    }
+  });
+  return tasks;
+}
+----
+
+NOTE: In the above example some deduplication is performed to ensure the same task is not returned to the UI when it is
+already present.
+
+We can then use this query in our application to subscribe to changes so that the front end is always updated when new
+data is returned from the server.
+
+[source,javascript]
+----
+this.tasks = [];
+this.getTasks().subscribe(result => {
+  this.tasks = result.data && result.data.allTasks;
+})
+----
+
+Note that it is also a good idea to unsubscribe from a query upon leaving a page. This prevents possible memory leaks.
+This can be done by calling unsubscribe() as shown in the following example. This code should be placed in the appropriate place.
+
+[source, javascript]
+----
+this.getTasks().unsubscribe();
+----
+
+== Client Network State Changes
+
+When using subscriptions to provide your client with realtime updates it is important to monitor network state. This is
+because of the architecture of websockets as described xref:sync-js-client-realtime-updates[above]. The problem is that if the
+client goes offline for any amount of time and some updates occur on the server in this time then the
+client will be out of sync. To remedy this {sync-client} provides a `NetworkStatus` interface which can be used along with
+the `NetworkInfo` interface to implement custom checks of network status. For more information about how to import and configure a
+custom network status checker please see xref:sync-js-client-advanced-topics[Advanced Topics].
+
+An example of using the provided Cordova Network Status, which is an implementation of the NetworkStatus interface can be seen below.
+This code should be used to re-run a query upon the client coming back to an online state.
+[source, javascript]
+----
+    import { CordovaNetworkStatus, NetworkInfo } from '@aerogear/voyager-client';
+    const networkStatus = new CordovaNetworkStatus();
+    const tasks = []
+
+    networkStatus.onStatusChangeListener({
+      onStatusChange(networkInfo: NetworkInfo) {
+        self.online = networkInfo.online;
+        if (self.online) {
+          this.tasks = return client.watchQuery({
+            query: GET_TASKS
+          });
+        }
+      }
+    });
+
+----
+
+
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,63 @@
 = Realtime Updates
 
-TODO
\ No newline at end of file
+Subscriptions are a core concept of GraphQL and one which allow us to enable real time updates and synchronisation of data.
```
wtrocki: I will remove `synchronization of data`. Subscriptions are sadly one way only. 
There are concepts in community for two way subscriptions (using socket for queries/mutations) so it is best to make it clear that we support one way only.

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,63 @@
...
+GraphQL Subscriptions use web sockets to enable clients to subscribe to published changes. {sync-client} comes with websocket and therefore
```
wtrocki: comes with WebSocket Link in Voyager Client

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,63 @@
...
+
+const client = createClient(config)
+```
+
+Once the client is configured we can then make use of subscriptions as follows.
+
+```
+
+getTasks() {
+  const tasks = client.watchQuery<Tasks>({
```
wtrocki: LOL. This is watching cache changes and it could be confusing for people. 
I would say let's remove it.

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,63 @@
...
+      query: GET_TASKS,
+      fetchPolicy: 'cache-first',
+      errorPolicy: 'none'
+  });
+
+  tasks.subscribeToMore({
+        document: TASK_MUTATED_SUBSCRIPTION,
+        updateQuery: (prev, { subscriptionData }) => {
```
wtrocki: This is specific aproach we use inside our app. It will require explanation before showing code.
We can leave update query empty and have chapter on this?

modules/ROOT/pages/_partials/sync/server-realtime-updates.adoc
```diff
@@ -1,3 +1,125 @@
 = Realtime Updates
 
-TODO
\ No newline at end of file
+A core concept of GraphQL is subscriptions. These subscriptions can be leveraged to provide a mechanism for real time updates.
```
wtrocki: Subscriptions `are` core concept of GraphQL that provide a mechanism..

modules/ROOT/pages/_partials/sync/server-realtime-updates.adoc
```diff
@@ -1,3 +1,125 @@
...
+To enable this on the server, an appropriate schema should be defined. A recommended implementation pattern can be seen in the schema below.
```
wtrocki: `them` ?

modules/ROOT/pages/_partials/sync/server-realtime-updates.adoc
```diff
@@ -1,3 +1,125 @@
...
+
+```
+const { PubSub } = require('apollo-server');
```
wtrocki: This is too verbose and as reader I have no clue what I should be focused on.
We have now nice separation in the showcase server. 
From the docs I get that there is some feature offered by voyager but we use apollo server. If we make that clear it will be easier for reader to get more info about subscriptions in google.

modules/ROOT/pages/_partials/sync/server-realtime-updates.adoc
```diff
@@ -1,3 +1,125 @@
...
+    },
+    Subscription: {
+        tasks: {
+            subscribe: () => pubSub.asyncIterator('tasks')
+        }
+  },
+}
+```
+
+In this example schema, we first set up a `Task` type, and then define the Mutations and Subscriptions required.
```
wtrocki: This is explaining our aproach for subscriptions. Maybe separate chapter.

modules/ROOT/pages/_partials/sync/server-realtime-updates.adoc
```diff
@@ -1,3 +1,125 @@
...
+When a new task is created, the `publish` method is called with information about the type of action performed along
+with the result of this action. The benefit of this type of subscription is that we can define one subscription channel
+`tasks` instead of `taskCreated`, `taskUpdated` etc. This reduces overhead on the client as there is only one
+subscription per object type. The client can then use the `action` field to decide what to do with subscription data.
+For documentation on how to use Subscriptions on your client,
+see xref:sync-js-client-realtime-updates[Realtime Updates]. You can then use the above schema as follows in a server.
+
+
+```
+const { VoyagerServer } = require('@aerogear/voyager-server')
```
wtrocki: This code is very volatile. I will avoid putting lots of code without adding much context - it will be maintenance hell in long run. I can tell that this got outdated already.

modules/ROOT/pages/_partials/sync/server-realtime-updates.adoc
```diff
@@ -1,3 +1,125 @@
...
+    schema: server.schema
+  }, {
+    server: httpServer,
+    path: '/graphql'
+  })
+})
+
+```
+
+NOTE: If using authentication on your server the subscription server should make use of the `onConnect` function to validate
```
wtrocki: I believe that this should be separate chapter. Really important info.

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
wei-lee: maybe it's a good idea to explain what subscriptions can do and can't do, to make it clear what use cases it supports.

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,63 @@
...
+An example of configuring the client for subscriptions can be seen below.
+
+```
+const config: DataSyncConfig {
+    ...
+    wsUrl: { WEB_SOCKET_URL_HERE }
+    ...
+}
+```
+Where wsUrl is the full URL of the websocket endpoint of your GraphQL server. Once this object is created, use it to initialise {sync-client} with:
```
wei-lee: this is about manually setting the `wsUrl`, right? Can we also mention that if you are using `mobile-services.json` file, it is setup automatically?

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,63 @@
...
+            const { id } = subscriptionData.data.tasks.task;
+            const filteredTasks = prev.allTasks.filter(item => {
+              return Number(item.id) !== Number(id);
+            });
+            prev.allTasks = filteredTasks;
+            return prev;
+          }
+        }
+      });
+}
```
wei-lee: are we missing a `return statement` here?

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
wtrocki: +100. Plus mention how other functionalities we provide interact with subscriptions.
How offline/reconnect stuff works etc.

StephenCoady:
Perfect. Thanks for the feedback @wtrocki and @wei-lee. I will make these changes straight away.

StephenCoady:
@wtrocki @wei-lee I made the changes you requested if you wouldn't mind taking another look. fyi I moved the protecting the subscription server part to the keycloak section as it requires set up and then cross-referenced in both places. can move it back if you think it fits better/worse.

I can't see the netlify preview for some reason, not sure if you guys can.

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,77 @@
...
+        return allTasks: [
+            ...prev,
+            newItem
+          ];
+        } else {
+          return prev
+        }
+    }
+  });
+  return tasks;
```
wei-lee: do we need to add some comments in the code to explain what the code does?

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,77 @@
...
+```
+
+This `getTasks()` function can then be used from a client service to control and view the updated tasks recieved
+from the server. An example can be seen below.
+
+```
+getTasks().subscribe(result => {
+  this.items = result.data.allTasks
+}
+```
```
wei-lee: are we missing some content on the client? For example, is there anything developers need to do when the app is online from the offline state? We said at the beginning that the client will miss all the subscription messages when it's offline, so when it's online again, should user run some queries again to get the data they probably have missed?

wei-lee: Another question I think worth a bit more explaining is the cache & subscription. Is it handled automatically by the voyager server, or do developers need to handle that?

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,77 @@
...
+To do this GraphQL Subscriptions utilise websockets to enable clients to subscribe to published changes.
+
+The architecture of websockets is as follows:
+
+* Client connects to websocket server
+* Upon certain events, the server can publish the results of these events to the websocket
+* Any _currently connected_ client to that websocket will receive these results
+* The client can close the connection at any time and will no longer recieve updates
+
+Websockets are not however, a queueing system. For example if the client is offline when the server publish the changes
```
wtrocki: I always prefer positive messages rather than negation. So instead of websockets are not.. we can say: "Websockets are perfect solution for delivering messages to currently active clients". Then we can follow up - for sending updates to non active clients we recommend push notifications. Mentioning queue here is not going to increase understanding as users can actually implement queue with changesets if they want to. So what we explain here is that we took upstream solution. Tested it with our features. Added auth and provide reference implementation if you looking for inspiration.

This is more advice rather than pointing some problem.

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,77 @@
...
+
+== Setting up a Client to use Subscriptions
+
+To set up a client to use subscriptions simply provide a wsUrl in your `mobile-services.json` file, {sync-client} will automatically
+apply the necessary logic to handle websocket requests. This can also be provided manually as seen below.
+
+```
+const config: DataSyncConfig {
+    ...
+    wsUrl: { WEB_SOCKET_URL_HERE }
```
wtrocki: This looks invalid. It should be string

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
wtrocki: We need to say why we use subscribeToMore over regular subscription (cache updates).
And entire pattern when using subscriptions:

1) Make network query to get fresh data
2) Watch cache changes
3) Subscribe
4) Unsubscribe when leaving view (important as can lead to memory leaks).
We do not do that because we had single view, now we have entire app so we may do that as well to not subscribe many times.

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
wtrocki: +1

modules/ROOT/pages/_partials/sync/server-auth.adoc
```diff
@@ -51,6 +51,29 @@ const server = VoyagerServer(apolloConfig, voyagerConfig)
...
+server.applyMiddleware({ app })
+  httpServer.listen({ 4000,
+}, () => {
+  new SubscriptionServer ({
+    execute,
+    subscribe,
+    schema: server.schema,
+    // add the onConnect method to check the client's token
+    onConnect: async connectionParams => {
+      return await keycloakService.validateToken(connectionParams)
```
wtrocki: It is important that we read all code samples line by line. 
In this case `keycloakService` was not defined and as reader this will confuse me a lot.

I personally prefer to have smaller code samples with something like:
```
onConnect: async connectionParams => {
      return await keycloakService.validateToken(connectionParams)
}
```
And then explain what this does and where needs to be put

StephenCoady: @wtrocki based on other comment I removed excess code and l stated above where the keycloakService was set up. But the way you have suggested is better and I will remove almost everything from this code snippet.

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
StephenCoady: @wei-lee to your first point, I can state that we have a network interface available in the sdk and that the developer can use that to detect network changes and therefore run some queries. I'm not sure how deep into that implementation I should go though - a lot of it will be very app specific. Wrt the cache - the developer will need to do some deduplication. We have an example of how we handle this but I am hesitant to put that code in the docs as again, its probably implementation specific. What do you guys think?

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,77 @@
...
+the client will not receive them. To receive updates the client must be currently connected to the websocket server,
+updates made over this websocket while the client is offline will not be replayed.
+{sync-client} comes with subscription support out of the box including auto-reconnection upon device restart or network reconnect.
+To enable subscriptions on your client set the following
+paramater in the {sync-client} config object. A DataSyncConfig interface is also available from {sync-client} if you wish to use it.
+
+== Setting up a Client to use Subscriptions
+
+To set up a client to use subscriptions simply provide a wsUrl in your `mobile-services.json` file, {sync-client} will automatically
+apply the necessary logic to handle websocket requests. This can also be provided manually as seen below.
```
wei-lee: actually, just look at the showcase app and I don't think this is the case. It looks like developer will need to pass the config to the sync client. See https://github.com/aerogear/ionic-showcase/blob/master/src/app/services/voyager.service.ts#L72

modules/ROOT/pages/_partials/sync/js-client-realtime-updates.adoc
```diff
@@ -1,3 +1,139 @@
...
+a `DataSyncConfig` interface available to use. {sync-client} will automatically
+apply the necessary logic to handle websocket requests.
+
+```
+const config: DataSyncConfig {
+    ...
+    wsUrl: "ws://your_websocket_url"
+    ...
+}
+```
```
wei-lee: all our example code should be in javascript

PR #623, opened 2019-02-13
Sync data access content

wei-lee:
## Motivation
Add content for the sync data access section. Also update the conflict resolution section to fix a few small issues.


modules/ROOT/pages/_partials/sync/server-data-access.adoc
```diff
@@ -1,3 +1,125 @@
 = Data Access
 
-TODO
\ No newline at end of file
+It is very likely a {sync-service} application will need to access data from a database for it to work. The GraphQL spec or Voyager libraries does not require an application to use a particular type of database, and the developer of the application can choose a database that is most suitable for the application.
```
finp: Is the point here that:

Typically when working with GraphQL, you use a database  to store data, for example, MongoDB or PostgreSQL.
To summarize the tasks associated with using a database from {sync-service}:

modules/ROOT/pages/_partials/sync/server-data-access.adoc
```diff
@@ -1,3 +1,125 @@
...
+
+. Initialize the database connection on app start
+. Add the database connection to the `context` object
+. Retrieve/persist data to the database in the application's resolvers
+
+== Initialize Database Connection
+
+Generally speaking, the application should set up the database connection on start, and fail if the connection can not be established. In order to connect to the database, usually the information like the hostname and port number of the database server, the credentials to authenticate etc will be required. It is recommended to use environment variables to set these details. 
+
+If the application is going to be deployed on OpenShift as described in the <<Running {sync-service} on OpenShift>> section, then the following environment variables will be setup for the application automatically:
```
finp: so, these env vars aren't available when doing local dev?
Would it be typical that user *doesn't* hook up voyager server to db during local dev times?

modules/ROOT/pages/_partials/sync/server-data-access.adoc
```diff
@@ -1,3 +1,125 @@
...
+
+const db = knex({
+  client: 'pg',
+  connection: dbConfig
+})
+----
+
+== Add Database Connection to `context` object
+
+Once the database connection is established, it needs to be accessible from a resolver so that it can be used to retrieve or persist data. The recommended way to achieve this is to add the database connection to the `context` object.
```
finp: kinda think this para is unnecessary

modules/ROOT/pages/_partials/sync/server-offline-and-conflict.adoc
```diff
@@ -114,53 +114,61 @@ Please see the chapter below for individual implementations.
...
 
-In the example above, the approach was to use `conflictHandler.resolveOnClient`. The server returns a special response to the client that contains the conflicting data and the client can make a choice on how to resolve it.
+In the example below, the approach was to use `conflictHandler.resolveOnClient`. When there is a conflict, the server returns a special response to the client that contains the conflicting data and the client can make a choice on how to resolve it. Since it will resolved on the client, it is not required to persist the data. However, if there is no conflict, the data sent by the client should be persisted.
 
 [source,javascript]
 ----
-// Data
-const serverState = ...
+// Read data from the data source, e.g. database
+const serverState = readData()
```
wtrocki: `readData()` should be in line 139

modules/ROOT/pages/_partials/sync/server-offline-and-conflict.adoc
```diff
@@ -114,53 +114,61 @@ Please see the chapter below for individual implementations.
...
+    await saveData(serverState)
     return serverState
 }
 ----
 
 == Resolving Conflicts on the Server
 
+In the example below, the approach was to use `conflictHandler.resolveOnServer`. When there is a conflict, the server will try to resolve it, using the specified resolution strategy. Once the conflict is resolved, the resolved data should be persisted. 
+
+Even though the conflict is resolved on the server, it did occur and the client should be notified about it in case it wants to do something about it. The server will return a message to the client to let it know that a conflict has happened, but it's resolved on the server, along with the resolved data.
```
wtrocki: ` in case it wants to do something about it` =>`notify user that sent data might have changed.`

modules/ROOT/pages/_partials/sync/server-data-access.adoc
wei-lee: no, these env vars won't be available locally unless the developer set them explicitly. 

No, that's not typical. Usually devs do hook up to db during local dev.

modules/ROOT/pages/_partials/sync/server-data-access.adoc
wei-lee: why?

modules/ROOT/pages/_partials/sync/server-offline-and-conflict.adoc
wei-lee: is that the only possible use case? I am trying to be generic here, not to suggest a particular use case.

modules/ROOT/pages/_partials/sync/server-data-access.adoc
wei-lee: kind of. My main point here is that the develop can choose whatever database that's suits their app.

modules/ROOT/pages/_partials/sync/server-offline-and-conflict.adoc
wtrocki: that is fine with me

modules/ROOT/pages/_partials/sync/server-data-access.adoc
finp: does mentioning Voyager in this context introduce the possibility that it doesn't work with sync service, ie will users care that it works with Voyager?

modules/ROOT/pages/_partials/sync/server-data-access.adoc
finp: so, i think the text is ambiguous here, what's the best way for me to propose edit, after merging this PR?

modules/ROOT/pages/_partials/sync/server-data-access.adoc
finp: or maybe reword, something like:

Add the database connection to the `context` object so that a resolver can access the database.

?

modules/ROOT/pages/_partials/sync/server-data-access.adoc
```diff
@@ -1,3 +1,125 @@
...
+|Database Server Port|DB_PORT
+|Database Server Username|DB_USERNAME
+|Database Server Password|DB_PASSWORD
+|Database name|DB_NAME
+|Use SSL|DB_SSL
+|====
+
+It is highly recommended that the application use these environment variables to setup the database connection.
+
+For example, an application needs to connect to a PostgreSQL database. The database, table and schema is already setup. On app start, the database connection can be established like this:
```
StephenCoady:
```suggestion
For example, if an application needs to connect to a PostgresQL database (where all tables and schemas are already set up) then a database connection can be established on app start as follows:
```

modules/ROOT/pages/_partials/sync/server-data-access.adoc
wei-lee: I don't think so. I think users do care, because some sync libraries work with only some type of database. For example, the old sync only works with mongodb. CouchDB only works with CouchDB etc. But with Voyager, you can use any.

modules/ROOT/pages/_partials/sync/server-data-access.adoc
wei-lee: create a new PR after this is merged?

modules/ROOT/pages/_partials/sync/server-data-access.adoc
wei-lee: sure

wei-lee:
@finp I have addressed most of the comments. Will it be ok if we merge this and improve it later on?

PR #625, opened 2019-02-18
Split up getting started 

finp:
## Motivation
The getting started is long an complex, reducing it to more manageable chunks should:
* make it easier for a new user
* make it possible to navigate to a specific section

## Tracked as
INTLY-130


finp:
@wei-lee  we discussed this last week

PR #628, opened 2019-02-18
fix: sync server docs fixes

darahayes:
## Description

* Minor fixes to the data access page
* Provide a lot more clarification to the sync server docs

### DIFF
--- a/modules/ROOT/pages/_partials/sync/server-data-access.adoc
+++ b/modules/ROOT/pages/_partials/sync/server-data-access.adoc
@@ -1,16 +1,16 @@
 = Data Access
 
-It is very likely a {sync-service} application will need to access data from a database for it to work. The GraphQL spec or Voyager libraries does not require an application to use a particular type of database, and the developer of the application can choose a database that is most suitable for the application. 
+A {sync-service} application will most likely need to access to a database. The GraphQL specification and {sync-server} modules not require an application to use a particular type of database, and the developer of the application can choose a database that is most suitable for the application. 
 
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/_partials/sync/server-data-access.adoc
```diff
@@ -64,12 +64,8 @@ const resolvers = //define the resolvers
...
   typeDefs,
   resolvers,
-  context: async ({ req }) => {
-    // pass request + db ref into context for each resolver
-    return {
-      req: req,
-      db: db.client
-    }
+  context: {
+    db: db.client
```
wtrocki: I think this may be just `db`as this comes from old format we had in server app:
https://github.com/aerogear/ionic-showcase/blob/master/server/src/index.js#L39

and above we have direct access to db.

modules/ROOT/pages/_partials/sync/server-offline-and-conflict.adoc
```diff
@@ -9,11 +9,21 @@ A *conflict* occurs whenever two or more clients try to modify the same data in
 
 In a GraphQL server, conflict detection and resolution happens exclusively in *mutations*.
 
-{sync-server} uses a concept of pluggable conflict resolution to allow developers to detect and handle conflicts regardless of the data storage.
+== Resolving Conflicts
+
+At a very high level, there is a typical flow to detecting and resolving conflicts.
+
+1. *A Mutation Occurs* - A client tries to modify or delete an object on the server using a GraphQL mutation.
```
wtrocki: 😍

PR #629, opened 2019-02-19
Sync getting started restructure

secondsun:
## Motivation
The docs here are a restructure of the presentation of the content to be more like guides than information dumps.  The work here is based on @finp's gist here : https://gist.github.com/finp/607f2f493c3a84f2c233a7127665ede8

## What
The docs are written to guide a user through local development then deploying to a cluster and finally binding the service and updating the client to use the bound service

## Why
The old docs are very information dense, but not very prescriptive.  The speak a lot to the "what" of sync but not the "how"

## Progress:

- [ x ] Server getting started
- [ x ] Client getting started
- [ x ] Creating the Server container
- [ x ] Deploying the server
- [ x ] Binding the server in MDC
- [ x  ] Updating the client


modules/ROOT/pages/_partials/sync/getting-started-mdc-client.adoc
```diff
@@ -0,0 +1,89 @@
...
+});
+----
+
+Once you have initialized the core, we can use it to configure the Data Sync client by setting the `openShiftConfig` property when we call `createClient`.
+
+.Updated data sync client
+[source,javascript]
+----
+let client = await createClient({
+    openShiftConfig:core.config
```
wtrocki:
```suggestion
   openShiftConfig: app.config
```

modules/ROOT/pages/_partials/sync/getting-started-openshift.adoc
```diff
@@ -0,0 +1,73 @@
+= Prerequisites
+* [x] You have Docker installed.
+* [x] You have access to an OpenShift cluster with the Service Catalog.
+* [x] You have completed the server getting started guide.
+
+= Overview
+
+To connect your Data Sync server and client to other mobile services, you need to run your application in OpenShift.  Data Sync provides a service catalog item to help with this.
```
wtrocki: >  you need to run your application in OpenShift.

Really strange to read this. We want to hide the fact that you could run this locally with no problem whatsoever?

wei-lee:
@secondsun can we merge this?

modules/ROOT/pages/_partials/sync/getting-started-openshift.adoc
```diff
@@ -0,0 +1,73 @@
...
+Data Sync requires your server application to be packaged as a Docker formatted container and published to a public respository such as link:https://hub.docker.com/[Docker hub]. 
+
+= Building and publishing the server container
+
+To build a server into a container, create a `Dockerfile` in the project's directory.  This container will need to include your server source code, its dependencies, and be configured to execute your server.
+
+As an example:
+
+[source,dockerfile]
+----
```
wtrocki: Why not use s2i node that will work here and it is fully supported way to run node.js app on OpenShift. 
There are also buckarest.gold images but that is hard to get by default on openshift.

modules/ROOT/pages/_partials/sync/index.adoc
```diff
@@ -8,9 +8,16 @@
 [id='{context}']
 = {sync-service}
 
-== Introduction
 
-include::../generic-service-intros.adoc[tags=sync-service]
+Unlike other AeroGear mobile services which provide a server and an API, Data Sync is a library that is used to develop services. Typically, you develop a {sync-service} service as follows:
```
wtrocki: `that can be used along with your node.js is application/backend` ?

wtrocki: `Data Sync` could be mapped to variable.

modules/ROOT/pages/_partials/sync/js-client-getting-started.adoc
```diff
@@ -1,50 +1,59 @@
...
-  strategies
-} from '@aerogear/voyager-client';
-----
+Below is a simple code example to connect to and query our {sync-server}.  To configure the client, a configuration object should be used. At a minimum, the `httpUrl` field is required and it is the url of the Voyager server application. If the client app uses subscription as well, then the `wsUrl` field is required too. 
 
-. Configuration
-+
-To configure the client, a configuration object should be used. At a minimum, the `httpUrl` field is required and it is the url of the Voyager server application. If the client app uses subscription as well, then the `wsUrl` field is required too. For example:
-+
+Core to connect to and query our server application would look like the following:
```
wtrocki: IMHO we could get that into the js-sdk sync/examples folder the same way we do with the server side examples. 
This way it will be closer to development and kept up to date. WDYT?

wtrocki:
Left couple comments but I think this is really good job.

modules/ROOT/pages/_partials/sync/getting-started-openshift.adoc
wei-lee: no, but this section is about getting started on openshift. The local dev part is covered in both data sync server and client sections

finp:
@secondsun @wtrocki  I think there are still issues in this, eg numbering
image

but I feel the best way forward is to merge this and handle the issues afterwards, wdyt?

PR #641, opened 2019-02-26
Edit to hello world example

finp:
Edit to hello world example


finp:
@darahayes got time to review?

PR #644, opened 2019-03-04
 fix(sync-server): update docs to reflect new reject method

darahayes:
Related Ticket: https://issues.jboss.org/browse/AEROGEAR-8679

### DIFF
--- a/modules/ROOT/pages/_partials/sync/server-offline-and-conflict.adoc
+++ b/modules/ROOT/pages/_partials/sync/server-offline-and-conflict.adoc
@@ -149,21 +149,19 @@ const { conflictHandler, strategies } = require('@aerogear/voyager-conflicts')
  {
    updateTask: async (obj, clientData, context, info) => {
      // 1. Read the Object from the database. This example uses Knex https://knexjs.org/.
-     const task = wait context.db('tasks').select().where('id', clientData.id).then((rows) => rows[0])
+     const task = await context.db('tasks').select().where('id', clientData.id).then((rows) => rows[0])
 
      // 2. Conflict Detection using `VersionedObjectState`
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/sync/server-offline-and-conflict.adoc
```diff
@@ -188,17 +186,30 @@ const { conflictHandler, strategies } = require('@aerogear/voyager-conflicts')
 
 When there is no conflict, `conflictHandler.nextState(clientData)` is called and the data is persisted. When a conflict occurs, the following happens.
 
-* `conflictHandler.resolveOnServer` is called with the `serverWins` strategy. In this case, the `resolvedState` will be the current server record with an updated version number. *The resolvedState must be persisted as the version is updated.*
+* `conflictHandler.resolveOnServer` is called with the `clientWins` strategy. In this case, the `resolvedState` will be the new data provided by the client. The newly `resolvedState` should be persisted.
```
finp: can we avoid 'should'?
How about 'You need to persist the `resolvedState` data.' or similar?

finp: @darahayes  merged as i posted review, just wondering tho, is my understanding correct in the comment I made?

darahayes: @finp apologies for that. Your understanding is definitely correct. How about I just switch the word `should` with `must`. **The newly `resolvedState` must be persisted**. WDYT?

finp:
@darahayes  still very passive, I think what you want the user to do is write code to persist the data, right?
If so, let's tell them that, and avoid ambiguity.

PR #647, opened 2019-03-05
feat: initial setup (AEROGEAR-8374)

laurafitzgerald:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8374

## What
See Progress

## Why
This is the first of the docs changes for the Mobile Security Service so this pr contains additional structure changes

## Verification Steps

_Steps_
1. Run the site locally using `npm run serve`
2. Navigate to the site in a browser

_Expected Result_
1. Navigation should be working
2. Page should appear as below:
screenshot 2019-03-05 at 16 19 09

## Checklist:

- [x] Code has been tested locally by PR requester
- [x] Changes have been successfully verified by another team member 

## Progress

- [x] Set up the initial structure
- [] Add details of landing page

## Additional Notes

Adding the full team for review as it would be good to get familiar with this repo and the structure. It's not required that everyone review the code but feel free to take a look.


laurafitzgerald:
thanks @davidffrench for the quick review @finp could I get a review from you. I noticed that there are a few places where files can be included from e.g. partials/generic-provisioning vs partials/keycloak/provision-the-idm
If I've used the incorrect file, just throw something across the canteen at me at some stage. 😂

laurafitzgerald:
@finp just seeing the netlify preview now. Does that mean for verification we don't need to do it locally? 😄

finp:
> @finp just seeing the netlify preview now. Does that mean for verification we don't need to do it locally?

yes, if the preview looks good, there is no need to run antora locally

modules/ROOT/nav.adoc
```diff
@@ -70,11 +70,15 @@ include::{partialsdir}/attributes.adoc[]
 *** xref:data-sync.adoc#auth[Authentication for {sync-service} Service]
 *** xref:data-sync.adoc#authorization[Authorization for {sync-service} Service]
 
-** xref:device-security.adoc[{device-security-service}]
-*** xref:device-security.adoc#introduction[Introduction]
-*** xref:device-security.adoc#setup[Setting Up]
-*** xref:device-security.adoc#device-trust[Performing Device Trust Checks]
-*** xref:device-security.adoc#cert-pinning[Certificate Pinning]
+** Security
```
finp: This should be a link, maybe that's where we explain the diff twixt security service and device security?

modules/ROOT/pages/security-service.adoc
```diff
@@ -0,0 +1,23 @@
...
+== Prerequisites
+
+* You are running OpenShift with {product-name} as described in xref:getting-started.adoc[Setting up AeroGear {product-name} on OpenShift].
+* You have provisioned Mobile Developer Console as described in link:{installation-link}[Provisioning Mobile Developer Console].
+
+[[setting-up-the-mobile-security-service]]
+== Setting Up the {mobile-security-service} Service
+
+// start {partialsdir}/generic-provisioning.adoc[]
+include::_partials/generic-provisioning.adoc[]
```
finp: As part of productization, we've removed some of the nested includes, (rats nest of includes). Turns out we don't have a very generic provisioning process. I'd recommend that you create a new file here, check IDM for recommended file layout

modules/ROOT/nav.adoc
finp: let me handle this as a separate pr

modules/ROOT/pages/security-service.adoc
finp: let's handle this after

PR #650, opened 2019-03-11
Edit for style and for productization

finp:
Trying to make the order of material more consumable


finp:
@wei-lee Other merged PRs have turned this into more of an edit for style, but if you have time for a review?

modules/ROOT/pages/data-sync.adoc
```diff
@@ -2,8 +2,38 @@
...
+[id='{context}']
+= Developing a {sync-service} App
+:toc:
+
+
+
+Unlike other AeroGear mobile services which provide a server and an API, Data Sync is a library that is used to develop services. Typically, you develop a {sync-service} service as follows:
+
+* Design a link:https://graphql.org/learn[GraphQL] schema
+* Develop a server side service and mobile application
```
wei-lee: should this just be `Develop a server side service` and the mobile application would be a separate step?

PR #651, opened 2019-03-11
Updating cordova docs

ziccardi:
# Motivation
Issue: [AEROGEAR-8066](https://issues.jboss.org/browse/AEROGEAR-8066)

### DIFF
--- a/modules/ROOT/pages/_partials/keycloak/setting-up-the-idm-sdk.adoc
+++ b/modules/ROOT/pages/_partials/keycloak/setting-up-the-idm-sdk.adoc
@@ -70,11 +70,20 @@ Cordova::
 +
 --
 // end::excludeDownstream[]
-Install the link:https://www.npmjs.com/package/@aerogear/auth[AeroGear Auth] package from link:https://www.npmjs.com/[NPM^]:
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/_partials/keycloak/setting-up-the-idm-sdk.adoc
```diff
@@ -70,11 +70,26 @@ Cordova::
...
 $ npm install @aerogear/auth
 ----
+. Install the `cordova-plugin-inappbrowser` plugin
++
+[source,bash]
+----
+$ cordova plugin add cordova-plugin-inappbrowser
+----
+. Use e.g `browserify` so that you can use the `require` statement
+. In the `init` snippet use `require` instead of `import`
```
jhellar: Maybe let's remove the above 2 lines, user will probably know how to import libraries and borwserify is probably not preferred way.

ziccardi: 👍

ziccardi:
@finp I believe they are unrelated: this is not about openshift, but for dealing with services using self signed certificates.

PR #657, opened 2019-03-22
add intro for data sync

finp:
## Motivation
add intro for data sync


finp:
@darahayes  Feel free to comment after the fact

PR #679, opened 2019-04-24
Sync templates chapter

wtrocki:
## Motivation

Add chapter listing templates.
Include different information for RHMI and upstream.

https://issues.jboss.org/browse/AEROGEAR-8976


modules/ROOT/pages/_partials/sync/openshift-templates.adoc
```diff
@@ -0,0 +1,29 @@
+
+= Provisioning the {sync-service} server applications using templates
+
+{sync-service} offers following OpenShift templates
+that will help developers with provisioning their DataSync applications to OpenShift platform.
+
+Templates:
+
+- DataSync Gateway
```
finp: could I use 3scale for this?
just trying to provide some context, eg 
This template provides a simple service to redirect APIs to different servers on OpenShift

wtrocki: Right. I might remove gateway thing as it is not related to the 3scale. It will be more like concept of the cloud app from RHMAP 3.x that provides way for various RHMAP services or even extenrnal customer backend. 

Calling it Mobile API gateway is risky as people may think 3scale instantly where this is just sever that people write themselves to connect to different services using GraphQL. Any suggestions welcome

ping @darahayes

wtrocki: I think I will extend information to make sure that thsi is an gateway app and not `Api management in any sort`

wtrocki: I think I will extend information to make sure that this is an gateway app and not `Api management in any sort`

darahayes: I think maybe there is some misunderstanding about 3Scale. 3Scale is a technology that does API Management. Coming from the website some of the features are:

* Rate limiting and Access Control
* Security
* Analytics
* Tools for Monetization (e.g. pay per requests)

In a strict sense, 3scale is not an API Gateway. The definition of an API Gateway [from NGINX](https://www.nginx.com/learn/api-gateway/) (who do great blogs and resources about microservices architectures) is the following:

> The API Gateway takes all the requests from the clients, then routes them to the appropriate microservice with request routing, composition, and protocol translation

Data Sync / GraphQL in general can be used to implement an API Gateway that would perform protocol translation to expose various aspects of the services in your backend (SQL, REST, RPC, SOAP, etc) and it all gets exposed via a GraphQL interface.

I think it's possible to confuse the two because both 3Scale and your Data Sync Service 'sit in front of your backend' but in completely different ways.

In fact you could end up putting 3scale in front of your Data Sync Service which would provide rate limiting, analytics, monetization, etc. Does that make sense?

wtrocki: Perfect explanation Dara. Thank you so much!

wei-lee: @wtrocki @finp are we sure it's a good idea to mention `API gateway` here? I would prefer we not to do that as I am sure it will cause some issues when 3scale sees this. Can we not just call it `Data Sync App`?

finp: @wtrocki @wei-lee  just an FYI re 3scale terminology and the potential for confusion:
https://www.redhat.com/en/resources/3scale-deploying-on-openshift-ebook

wtrocki: Yeah. I have created separate PR to change name.

PR #695, opened 2019-05-17
Data sync reorg https://issues.jboss.org/browse/AGMS-330

finp:
## Motivation
Data sync docs are long and complex, and difficult to review, consume and edit
## What
break out each feature to make it clearer, 
move files to module


wei-lee:
+1 on the new structure. I think it is much better than the previous one, and make things easier to find for developers.

finp:
@wei-lee could you take a look again? and consider https://issues.jboss.org/browse/AEROGEAR-9247

wei-lee:
@finp is there anything particular I should check? From the preview it looks like the query page is still empty?

modules/ROOT/pages/_partials/data-sync/js-client-query.adoc
```diff
@@ -0,0 +1,123 @@
...
+// For our client application, we will connect to the local service.
+let config = {
+  httpUrl: "http://localhost:4000/graphql",
+  wsUrl: "ws://localhost:4000/graphql",
+}
+
+async function queryPeople() {
+
+  // Actually create the client
+  let offlineClient = new OfflineClient(config);
```
wei-lee: this will be different if the app is using `mobile-services.json` file. We should given an example for how to use that file to init the client.

finp: +1 will handle in https://issues.jboss.org/browse/AEROGEAR-9303

PR #698, opened 2019-05-23
AGMS-287 resolve adoc issue

finp:
## Motivation
fix rendering issue 

## What
masquerade a block as described in https://asciidoctor-docs.netlify.com/asciidoc/1.5/blocks/open/
## Why
fix issue 
## How
labeling the block seems to fix issue


jstaffor:
Screenshot from 2019-05-24 09-23-10

Good job @finp :+1: 

I'll add the fix to master in the downstream repo and publish to get the fix out :+1:  (This fix will be applied anytime someone pulls from upstream so it's all future proof)

PR #716, opened 2019-07-04
Subscriptions docs1

finp:
## Motivation

Edit Dara's PR to make it more consumable

## What
* introduce concepts
* provide procedures
* clearly delineate options for user

### DIFF
--- a/modules/ROOT/pages/_partials/data-sync/realtime-intro.adoc
+++ b/modules/ROOT/pages/_partials/data-sync/realtime-intro.adoc
@@ -1,0 +1,30 @@
+[id="realtime-intro-{context}"]
+= Introduction to real-time updates
+
+After you have developed some queries and mutations, you might want to implement real-time updates.
+These are supported in the GraphQL specification by an operation type called `Subscription`.
+To support scalable, reliable subscriptions in a production environment, Data Sync implements subscriptions using an MQTT PubSub subscription mechanism, however you might want to use the Apollo PubSub module to develop proof-of-concept applications.  
+
+When coding for real-time updates, you use the following modules:
+
+* @aerogear/voyager-server - supports clients that use voyager-client to enable GraphQL queries and mutations
+* @aerogear/voyager-subscriptions - supports clients that use voyager-client to enable GraphQL subscriptions
+* @aerogear/graph-mqtt-subscriptions - supports GraphQL resolvers connections to a MQTT broker
+
+GraphQL Subscriptions utilise websockets to enable clients to subscribe to published changes.
+
+The architecture of websockets is:
+
+* Client connects to websocket server.
+* Upon certain events, the server can publish the results of these events to the websocket.
+* Any _currently connected_ client to that websocket receives these results.
+* The client can close the connection at any time and no longer receives updates.
+
+To receive updates the client must be currently connected to the websocket server, updates made over this websocket while the client is offline are not consumed by the client.
+To support inactive clients, use Push Notifications.
+
+{sync-client} comes with subscription support out of the box including auto-reconnection upon device restart or network reconnect.
+
+.Additional resources
+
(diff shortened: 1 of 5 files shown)


modules/ROOT/pages/_partials/data-sync/realtime-intro.adoc
```diff
@@ -0,0 +1,30 @@
...
+These are supported in the GraphQL specification by an operation type called `Subscription`.
+To support scalable, reliable subscriptions in a production environment, Data Sync implements subscriptions using an MQTT PubSub subscription mechanism, however you might want to use the Apollo PubSub module to develop proof-of-concept applications.  
+
+When coding for real-time updates, you use the following modules:
+
+* @aerogear/voyager-server - supports clients that use voyager-client to enable GraphQL queries and mutations
+* @aerogear/voyager-subscriptions - supports clients that use voyager-client to enable GraphQL subscriptions
+* @aerogear/graph-mqtt-subscriptions - supports GraphQL resolvers connections to a MQTT broker
+
+GraphQL Subscriptions utilise websockets to enable clients to subscribe to published changes.
```
darahayes: I might suggest something like `GraphQL Subscriptions enable clients to subscribe to events <generated within|that happen on> the server over a websocket connection`

modules/ROOT/pages/_partials/data-sync/realtime-intro.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+The architecture of websockets is:
```
darahayes: This title is a little inaccurate and possibly confusing. Websockets is a lower level communication protocol, over which subscriptions are delivered.

I would maybe avoid using the term `websocket server` too because it's possible that someone would read that as a different server to the data sync server (which is not the case).

I would maybe change the title here to `The architecture of subscriptions is` or `The general flow for subscriptions is` or something like that.

As for the bullet points themselves, I got a strong impression here that any and all clients receive every single event that happens on the server, which is not the case. As part of the initial connection, the client **Subscribes** to particular events it is interested in. (example in a chat app: I want to know about messages only in a particular channel. Not all messages in the whole universe). I might go with something like this

```
* Client connects to the server using websockets, and subscribes to certain events
* Upon events, the server notifies the clients that are <subscribed to|interested in> those events
* Any _currently connected_ client that is subscribed to a given event receives it.
* The client can close the connection at any time and no longer receives updates.
```

modules/ROOT/pages/_partials/data-sync/realtime-intro.adoc
```diff
@@ -0,0 +1,30 @@
...
+
+* Client connects to websocket server.
+* Upon certain events, the server can publish the results of these events to the websocket.
+* Any _currently connected_ client to that websocket receives these results.
+* The client can close the connection at any time and no longer receives updates.
+
+To receive updates the client must be currently connected to the websocket server, updates made over this websocket while the client is offline are not consumed by the client.
```
darahayes:
```suggestion
To receive updates, the client must be currently connected to the server, the client will not receive events from subscriptions while offline.
```

Mainly trying to avoid the sentence `updates made over this websocket` which is inaccurate.

modules/ROOT/pages/_partials/data-sync/realtime-intro.adoc
finp: updated

modules/ROOT/pages/_partials/data-sync/realtime-intro.adoc
finp: updated

modules/ROOT/pages/_partials/data-sync/realtime-intro.adoc
finp: updated

PR #717, opened 2019-07-08
fix gettingstarted files

finp:
## Motivation
reduce getting started, queries and mutations doc to minimal content


finp:
@StephenCoady I'm having problems getting the mutation to work, could I ask you to take a look?
(works in playground, but not in client app)

modules/ROOT/pages/_partials/data-sync/js-client-mutate.adoc
```diff
@@ -37,29 +28,30 @@ async function addPerson() {
   let client = await offlineClient.init();
 
   // Execute the mutation
-  client.query({
+  client.mutate({
       fetchPolicy: 'network-only',
```
StephenCoady: @finp I think this link might be your problem - probably left over from the query that was here. remove this and it should work.

jstaffor: @finp - I tested this locally and it works, :+1: @StephenCoady 

I also modified it to write to the html page using:

 document.write(JSON.stringify(data))

Screenshot from 2019-07-08 15-57-55

finp: @StephenCoady  thanks that was the trick. BTW we don't have a cli way to guide users thru this.
@jstaffor  I'm going with console cos a) we don't have a cli way of getting output b) we ask users later to use console https://docs.aerogear.org/aerogear/latest/ds-conflict.html#listening-to-conflicts

jstaffor: :+1:

PR #722, opened 2019-07-15
add index

finp:
## Motivation
This is a requirement for publishing docs as part of intly community, preview at scalecomputing.org


modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,3 @@
+= AeroGear Mobile Services
+
+See also: https://www.aerogear.org
```
jstaffor:
```suggestion
Also see https://www.aerogear.org
```

finp: will revise later

PR #726, opened 2019-07-19
Aerogear 9227 - Review supported configs (dev complete checkin)

jstaffor:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9227

## What
https://issues.jboss.org/browse/AEROGEAR-9227

## Why
https://issues.jboss.org/browse/AEROGEAR-9227

## How
https://issues.jboss.org/browse/AEROGEAR-9227


wtrocki:
Looks good. I will hold of with merge until we confirm this stack by testing.

modules/ROOT/pages/supported_versions.adoc
```diff
@@ -3,9 +3,10 @@ include::_partials/attributes.adoc[]
...
+As of July 2019, the following versions are tested and supported:
 
-* Cordova: version 8
-* Nodejs: version 8 and 10
-* Android SDK: API level 27
-* XCode: version 10.1
\ No newline at end of file
+* Cordova: version 9
+* Nodejs: version 10 and 12 LTS
+* Android SDK: API level 29
```
jhellar: Should this be 28? From what I've looked, 29 was not released yet.

modules/ROOT/pages/supported_versions.adoc
```diff
@@ -3,9 +3,10 @@ include::_partials/attributes.adoc[]
...
+Generally speaking, AeroGear Mobile Services aim to support the latest versions of tools and technologies.
+As of July 2019, the following versions are tested and supported:
 
-* Cordova: version 8
-* Nodejs: version 8 and 10
-* Android SDK: API level 27
-* XCode: version 10.1
\ No newline at end of file
+* Cordova: version 9
+* Nodejs: version 10 and 12 LTS
```
jhellar: Version 10 is now LTS. But should we mention it all? Would just `version 10 and 12` be ok?

modules/ROOT/pages/supported_versions.adoc
wtrocki: I remember @danielpassos trying our stuff on top of the beta release. 
I wanted to save another bump but we can downgrade to 28

modules/ROOT/pages/supported_versions.adoc
wtrocki: good point. Let's mention both as LTS

modules/ROOT/pages/supported_versions.adoc
danielpassos: Well, Android 29 still in [preview version](https://developer.android.com/preview)

modules/ROOT/pages/supported_versions.adoc
jstaffor: Food for thought:
Node v**10.0.0** (Current) - https://nodejs.org/fr/blog/release/v10.0.0/
Node v**10.13.0** (**LTS**) - https://nodejs.org/fr/blog/release/v10.13.0/

In terms of accurately presenting info, I think we need to state "Node v10.13.0 (LTS)" if we want to mention LTS as it seems to me that "Node v10" isn't LTS. I'm open to correction on this as I'm not an expert in this area. What are your thoughts on this?

modules/ROOT/pages/supported_versions.adoc
jstaffor: I think it's safer to state "Android SDK: API level 28". Does any know what version(s) did we test this on/plan to test on?

jhellar: Currently our integration tests run against API level 28, also device we are/will be testing on manually is 28.

wtrocki: Agree! 28 is better for now

modules/ROOT/pages/supported_versions.adoc
wtrocki: We will always work and test on LTS

modules/ROOT/pages/supported_versions.adoc
jstaffor: Great - I'll make the udpate :+1:

modules/ROOT/pages/supported_versions.adoc
jstaffor:
```suggestion
* Nodejs: version Node v10.16.0 (LTS) and Node v12.x
```

Looking at the versions - there's no LTS for Node v12 hence the proposal - open to suggestion.

finp: it's explained here https://endoflife.date/nodejs
v12 will become LTS in Oct (as far as I can see)

modules/ROOT/pages/supported_versions.adoc
```diff
@@ -3,9 +3,10 @@ include::_partials/attributes.adoc[]
...
+As of July 2019, the following versions are tested and supported:
 
-* Cordova: version 8
-* Nodejs: version 8 and 10
-* Android SDK: API level 27
-* XCode: version 10.1
\ No newline at end of file
+* Cordova: version 9
+* Nodejs: version Node v10.16.0 (LTS) and Node v12.x
+* Android SDK: API level 29
```
jstaffor:
```suggestion
* Android SDK: API level 28
```

finp:
@wtrocki @jhellar  do you think we're good to merge this yet?

wtrocki:
Yes

finp:
@jhellar We can handle the Xcode version separately: My question for you and @wtrocki  is what happens if someone uses Xcode 10.2? do things stop working? 
Would it make sense to say Xcode 10.2 or higher?

wtrocki:
@finp We should not care about this at all. Our relation to xcode comes with Cordova so let's say that we say that you should use Cordova 9 and this comes with specific support for android and IOS that has Xcode. We mentioning it for convenience so people do not need to go to cordova and check what versions they need. Generally, version range for support is very wide so I see this more as 
mentioning on what stack we are testing this.

PR #727, opened 2019-07-21
publish assemblies to netlify adoc

finp:


finp:
stopping netlify - not picking up changes

PR #743, opened 2019-07-25
AGMS-282 - Review all images prior to GA

jstaffor:
## Motivation
https://issues.jboss.org/browse/AGMS-282

## What
https://issues.jboss.org/browse/AGMS-282

## Why
https://issues.jboss.org/browse/AGMS-282

## How
https://issues.jboss.org/browse/AGMS-282


jstaffor:
I can't find any instances where the images are missing in the docs so I think we're good. Once merged, I'll rebuild the content downstream and CCUTIL should identify any images with missing images - I say this as I can rem times when it will no build due to missing images. If there such instances, I'll fix them upstream in a follow-on PR to this :+1:

PR #760, opened 2019-08-02
update upstream methodology

finp:


finp:
IDM review:
- no toc
- existing problem:
image

PR #764, opened 2019-08-07
chore: remove reference to old documentation

wtrocki:
## Motivation

Disable docs from rendering but keep it for reference just in case some downstream versioning etc.


finp:
conversation at https://issues.jboss.org/browse/AEROGEAR-9656

PR #768, opened 2019-08-08
AGMS-743 publish downstream

finp:
## Purpose
create docs for rhpds users
Type:
- [x ] Technical change
- [ ] Style edit


finp:
closing, I'm considering a different approach after some thinking

PR #772, opened 2019-09-11
AEROGEAR-9888-explain versions

finp:
## Purpose

add more context around versions, to avoid scenarios like:

I just released a new version of my app, and know there's an issue, but I can't disable it, because it's not appearing in MSS console

### DIFF
--- a/modules/ROOT/pages/_partials/app-security/using-the-admin-ui.adoc
+++ b/modules/ROOT/pages/_partials/app-security/using-the-admin-ui.adoc
@@ -11,9 +11,14 @@ The console is only available after the {app-security-service} Service has been 
 
 .Overview of the Mobile Security Console
 
-The {app-security-service} Console allows you to monitor applications, their respective versions, and disable versions of those applications.
+The {app-security-service} Console allows you to monitor mobile apps, deployed versions, and disable versions of those mobile apps.
 
-Opening the console lists all applications. Click on an application to view version details for that application.
+Opening the console lists all applications. Click on an application to view deployed versions of that mobile app.
+
+A deployed version is registered with Mobile Security after:
+
+* A version of the app is released, for example, 1.2.3.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/app-security/using-the-admin-ui.adoc
```diff
@@ -11,9 +11,14 @@ The console is only available after the {app-security-service} Service has been
 
 .Overview of the Mobile Security Console
 
-The {app-security-service} Console allows you to monitor applications, their respective versions, and disable versions of those applications.
+The {app-security-service} Console allows you to monitor mobile apps, deployed versions, and disable versions of those mobile apps.
 
-Opening the console lists all applications. Click on an application to view version details for that application.
+Opening the console lists all applications. Click on an application to view deployed versions of that mobile app.
+
+A deployed version is registered with Mobile Security after:
```
craicoverflow: Should this say "registered with the Mobile Security Service". I don't believe it's ever referred to as just "Mobile Security".

modules/ROOT/pages/_partials/app-security/using-the-admin-ui.adoc
```diff
@@ -11,9 +11,14 @@ The console is only available after the {app-security-service} Service has been
...
+
+* A version of the app is released, for example, 1.2.3.
```
craicoverflow: Using an unordered list suggests only **one** of these two events need to occur in order for a deployed version to be registered.

The actual case being that point 1 is a requirement for point 2 to be possible.

Do you agree? Maybe this should be an ordered list or some other alternative?

finp: I don't think an unordered  list implies OR, but lemme try to make it more explicit ;)

PR #784, opened 2019-09-24
Adds downstream mobile-clients-services-all-idm-provisioned image

briandooley:
review @finp please


finp:
something's messed up, see 
https://deploy-preview-784--mobile-docs.netlify.com/aerogear/latest/getting-started-configuring/

briandooley:
> something's messed up, see
> https://deploy-preview-784--mobile-docs.netlify.com/aerogear/latest/getting-started-configuring/

You mean down and upstream images appearing at the same time?  Yeah, I spent a good chunk of time looking into whether I had the tags wrong or something.  My guess is that there is some setting in netlify is not set correctly.  I tried playing with the include-tags value in netlify-build.sh but no joy.  Was going to wait until I could talk to you directly.  You ever see it before?

PR #786, opened 2019-09-24
Mobile clients service bind 1 down

briandooley:
@finp another review please


finp:
is it clear how to proceed after yesterday's chat?

briandooley:
yep, use tags={include-tags} to tell asciidoc to pay attention to upstream/downstream tag, as I recall.

briandooley:
@finp Any chance of approving this?  Its showing the upstream image on netlify so I reckon it works now.

PR #802, opened 2019-10-04
Aerogear 9757 bulk

briandooley:
@finp When you are ready.


finp:
didn't we agree that we only need one image? that whatever is current should 'win'?

briandooley:
@finp This was to get rid of the "double images" problem.  I have changed this again to leave things as they were before I changed them.

briandooley:
> the images are updated in a separate PR?

Is that your way of saying "You forgot to include in the images"?  ಠ‿↼

I've checked. All the image files referred to in this PR are already in the images directory in the master branch.  Will wait until I chat to you on Monday in case I am misinterpreting what you are saying.

PR #804, opened 2019-10-09
Added service worker instructions

ziccardi:
## Purpose
Added instruction about the service worker.

Type:
- [x] Technical change

### DIFF
--- a/modules/ROOT/pages/_partials/push/registering_device.adoc
+++ b/modules/ROOT/pages/_partials/push/registering_device.adoc
@@ -87,8 +87,9 @@ Web::
 ----
 import { PushRegistration } from "@aerogear/push";
 
+// Register your app for push notification, passing the relative URL to your service worker
 new PushRegistration(app.config)
-  .register({serviceWorker: 'your-service-worker.js'}).then(() => {
+  .register({serviceWorker: '/path/to/your-service-worker.js'}).then(() => {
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/push/registering_device.adoc
```diff
@@ -105,6 +106,39 @@ NOTE: Optionally, you can pass the parameters below to the register method
   timeout: 5000 // in milliseconds
 }
 ----
+
+WARNING: It is very important that you provide the correct path to your service worker: it will be responsible of
```
briandooley:
```suggestion
WARNING: It is very important that you provide the correct path to your service worker: it will be responsible for
```

finp: I'd suggest we try this rewrite again.

finp: I'd suggest removing this note, where can you provide an incorrect path in any software?

finp:
@danielpassos  that's fine.

modules/ROOT/pages/_partials/push/registering_device.adoc
danielpassos: I have fixed it directly on WebPush branch. See: https://github.com/aerogear/mobile-docs/commit/695d240578182ebc57f3f71d55c1c06608ac84b0

PR #807, opened 2019-10-14
AEROGEAR-10015- update to include new version

finp:
## Purpose
Add new version to provide fallback as new installation lands


joaedwar:
LGTM

